PR #91, opened 2026-03-28
Add pre-deploy backup of production site

### LINKED ISSUE #78: Add pre-deploy backup of production site
## Parent issue

Extracted from #45 (Add rollback capability) — Phase 1.

## Problem

Deployment uses \`rsync --delete\` to production. A bad deploy destroys the previous working version with no recovery path.

## Proposed solution

Add a step to \`deploy.yml\` before the rsync that creates a timestamped backup on the server:

\`\`\`yaml
- name: Backup current deployment
  run: |
    ssh -p 18765 user@host \\
      "cp -r www/deming.../public_html www/deming.../public_html.backup.\$(date +%Y%m%d%H%M%S)"
\`\`\`

Add a cleanup step that keeps only the last 5 backups to prevent disk bloat.

## Success criteria

- [ ] A timestamped backup is created before each deployment
- [ ] Backup is a complete copy of the production site
- [ ] Old backups are automatically cleaned up (keep last 5)
- [ ] Backup step runs in both deploy workflows

## Size: S

lddurbin:
## Summary
- Adds a timestamped server-side backup step before each `rsync --delete` deployment, preventing data loss from bad deploys
- Automatically cleans up old backups, keeping only the last 5
- Added to both `deploy.yml` and `deploy-optimized.yml` with cross-reference comments

Closes #78

## Test plan
- [ ] Trigger a deploy and verify a backup directory appears at `www/deming.leedurbin.co.nz/backups/`
- [ ] Trigger 6+ deploys and verify only the 5 most recent backups are retained
- [ ] Verify the deployed site works correctly after the backup step runs

### DIFF
--- a/.github/workflows/deploy-optimized.yml
+++ b/.github/workflows/deploy-optimized.yml
@@ -108,6 +108,21 @@ jobs:
           exit 1
         fi
         
+    # Keep in sync with deploy.yml
+    - name: Backup current deployment
+      run: |
+        DEPLOY_PATH="www/deming.leedurbin.co.nz/public_html"
+        BACKUP_DIR="www/deming.leedurbin.co.nz/backups"
+        SSH_CMD="ssh -p 18765 u197-gmrgybn3hkn2@ssh.leedurbin.co.nz"
+
+        echo "Creating pre-deploy backup..."
+        $SSH_CMD "mkdir -p ${BACKUP_DIR} && cp -r ${DEPLOY_PATH} ${BACKUP_DIR}/public_html.backup.\$(date +%Y%m%d%H%M%S)"
+        echo "Backup created"
+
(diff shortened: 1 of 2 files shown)


lddurbin:
Thanks for the review. Addressed in 7d5f947:

**First deploy guard** — Agreed. Added `if [ -d ... ]` check so the step skips gracefully when `public_html` doesn't exist yet.

**Single SSH connection** — Agreed. Combined backup + cleanup into one `$SSH_CMD` call.

**Duplicate code fragility** — Acknowledged as a known trade-off. The cross-reference comments are sufficient at this scale.

**`set -e` consistency** — Skipping this one. The `&&` chaining already propagates errors in the remote command, and the local script is now just variable assignment → one SSH call → echo. Adding `set -e` wouldn't change behaviour here.

PR #95, opened 2026-03-28
Pin all build tool versions for reproducible builds

### LINKED ISSUE #46: Pin all build tool versions for reproducible builds
### Description

Several build dependencies use floating versions, making builds non-reproducible:

| Component | Pinned? | Location |
|---|---|---|
| R packages | Yes (renv.lock) | `renv.lock` |
| R version | **No** — set to `'release'` | `deploy.yml:23` |
| Quarto version | Hardcoded to 1.4.549 | `deploy.yml:69` |
| Pandoc version | **No** — floating | `deploy.yml` (via setup-pandoc) |
| System libraries | **No** — `apt-get install` without versions | `deploy.yml:28–40` |

When R 4.5.0 ships, the CI environment silently upgrades, potentially breaking package installations or changing computation results.

**Related:** #22 covers pinning R version specifically. This issue is broader — all floating dependencies.

### Affected files

- `.github/workflows/deploy.yml:23,28–40,69`

### Recommended fix

```yaml
- uses: r-lib/actions/setup-r@v2
  with:
    r-version: '4.4.0'  # Match renv.lock

- uses: r-lib/actions/setup-pandoc@v2
  with:
    pandoc-version: '3.1.11'  # Pin to tested version
```

For system libraries, consider using a Docker base image with pre-installed dependencies to lock versions.

### Success criteria

- [ ] R version in CI matches `renv.lock` exactly
- [ ] Pandoc version is pinned
- [ ] Quarto version is reviewed and updated intentionally (not left stale)
- [ ] Two builds from the same commit produce identical output

lddurbin:
## Summary
- Pin R version to `4.4.0` (matching `renv.lock`) — was `'release'` which silently upgraded to 4.5.x
- Pin Pandoc to `3.1.11` — was floating with no version specified
- Replace hardcoded `R-4.5` library paths with dynamically derived paths
- Add R version to cache keys to prevent ABI-incompatible cache reuse
- Document Quarto 1.4.549 pin as intentional (already was pinned, just undocumented)
- Applied consistently across both `deploy.yml` and `deploy-optimized.yml`

Closes #46

## Test plan
- [ ] CI build passes with pinned R 4.4.0
- [ ] renv restore succeeds (no R version mismatch)
- [ ] Quarto render produces expected output
- [ ] Two builds from the same commit produce identical output

### DIFF
--- a/.github/workflows/deploy-optimized.yml
+++ b/.github/workflows/deploy-optimized.yml
@@ -19,7 +19,7 @@ jobs:
     - name: Setup R
       uses: r-lib/actions/setup-r@v2
       with:
-        r-version: 'release'
+        r-version: '4.4.0'  # Must match renv.lock R version
         use-public-rspm: true
         
     - name: Cache system dependencies
@@ -49,6 +49,8 @@ jobs:
 
     - name: Setup Pandoc
       uses: r-lib/actions/setup-pandoc@v2
+      with:
+        pandoc-version: '3.1.11'  # Pinned for reproducible builds
       
     - name: Cache Quarto installation
       id: cache-quarto
@@ -58,6 +60,7 @@ jobs:
         key: quarto-1.4.549
         
     - name: Install Quarto
+      # Pinned to 1.4.549 — last tested version. Upgrade intentionally, not automatically.
       if: steps.cache-quarto.outputs.cache-hit != 'true'
       run: |
         curl -LO https://github.com/quarto-dev/quarto-cli/releases/download/v1.4.549/quarto-1.4.549-linux-amd64.deb
@@ -70,8 +73,8 @@ jobs:
       uses: actions/cache@v4
       with:
         path: ~/.local/share/renv
-        key: ${{ runner.os }}-renv-${{ hashFiles('renv.lock') }}
-        restore-keys: ${{ runner.os }}-renv-
+        key: ${{ runner.os }}-r4.4.0-renv-${{ hashFiles('renv.lock') }}
+        restore-keys: ${{ runner.os }}-r4.4.0-renv-
 
     - name: Restore R packages
       shell: Rscript {0}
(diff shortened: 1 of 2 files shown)


lddurbin:
Thanks for the review — addressed in 121fe86:

1. **`substr` → `strsplit`**: Agreed, fixed. Using `strsplit(R.version$minor, "\\\\.")[[1]][1]` now handles two-digit minor versions correctly.

2. **Cache key drift**: Good catch. Both workflows now use `${{ steps.setup-r.outputs.installed-r-version }}` in cache keys instead of a hardcoded string. Added `id: setup-r` to `deploy-optimized.yml`.

3. **Stale Quarto comment**: Replaced with a dated "last reviewed" note: `(reviewed 2026-03)`.

4. **Hardcoded architecture string**: Respectfully pushing back here. GitHub Actions Ubuntu runners are exclusively x86_64, and this is unlikely to change without a major platform announcement. Adding dynamic architecture detection would mean an extra `Rscript` invocation per step for a scenario that won't arise on this CI platform. If the project ever moves to ARM runners, the renv library structure would need broader rework anyway. Happy to add a comment if you feel strongly, but I'd rather not over-engineer this.

lddurbin:
Addressed in f63bf4d:

1. **Pandoc review date**: Added `(reviewed 2026-03)` to the pandoc version comment in both workflows, matching the Quarto pin style.
2. **`actions/cache` version**: Upgraded `deploy.yml` from `@v3` to `@v4` to match `deploy-optimized.yml`.

lddurbin:
Good catch — confirmed the escaping bug. Four backslashes through YAML literal block → shell single quotes → R string parsing produced `\\.` (backslash + any char) instead of matching a dot. `strsplit("4.0", "\\.")` returns `"4.0"` unsplit, giving `R-4.4.0` instead of `R-4.4`.

Fixed in 0893619 using `fixed=TRUE` to avoid the escaping layers entirely.

PR #100, opened 2026-03-30
Extract duplicated R setup chunk into shared project-level config

### LINKED ISSUE #68: Extract duplicated R setup chunk into shared project-level config
## Problem

All 42 `.qmd` files contain an identical 2-line R setup chunk:

```r
knitr::knit_hooks$set(crop = knitr::hook_pdfcrop)
source(here::here("R/functions/main-functions.R"))
```

This is 84 lines of pure duplication. Any change to the setup (e.g. adding a new hook or sourcing a new file) requires editing every single chapter file. This also creates drift risk as Days 6-12 are converted.

**Note:** This is distinct from #48 (OJS interactive patterns) — this is about the R/knitr setup, not OJS boilerplate.

## Proposed solution

Use Quarto's project-level knitr configuration. Options (in order of preference):

1. **`_quarto.yml` execute options** — set `include-before-body` or knitr opts at project level
2. **Shared `_common.R`** — create an R script sourced via knitr's `source` option in `_quarto.yml`
3. **`.Rprofile` project hook** — source the setup in the project `.Rprofile` (less explicit)

Whichever approach is chosen, the setup chunk should then be removed from all 42 files.

## Success criteria

- [ ] R setup (knitr hooks + function sourcing) is defined in exactly one place
- [ ] All 42 `.qmd` files no longer contain the duplicated setup chunk
- [ ] `quarto render` succeeds with no regressions
- [ ] New chapter files created for Days 6-12 will not need the setup chunk

## Size: M

Requires identifying the right Quarto/knitr mechanism, implementing it, removing the chunk from 42 files, and verifying the build. Moderate token usage for research + bulk edits + build verification.

lddurbin:
## Summary
- Centralizes the duplicated 2-line R setup chunk (knitr crop hook + function sourcing) into `R/setup.R`, auto-sourced via `.Rprofile`
- Removes the setup chunk from all 42 `.qmd` files (−291 lines)
- Updates both templates and the conversion guide so new chapters won't include the chunk

Closes #68

## Test plan
- [x] `quarto render` succeeds with no errors
- [x] Spot-checked rendered HTML for R output (clocks, control charts)
- [x] Verified no setup chunks remain in `content/` files
- [ ] Verify deployed site renders correctly


lddurbin:
Thanks for the review — both points addressed in 4d37d5c:

1. **Silent failure message**: Added an `else` branch with `message()` so a missing `R/setup.R` is immediately visible in the console rather than surfacing as cryptic downstream errors.

2. **Dropped `here::here()` dependency**: Replaced with `file.path("R", "functions", "main-functions.R")`. Since `.Rprofile` always runs from the project root, `here` is unnecessary here — and this avoids loading both `here` and `knitr` namespaces in interactive sessions that don't need them.

Also added the full issue URL cross-reference to the `.Rprofile` comment block as suggested.

Build verified: `quarto render` succeeds with no regressions.

Issue #85, opened 2026-03-27
Add roxygen2 documentation to R helper functions


lddurbin:
## Parent issue

Extracted from #47 (Consolidate duplicated R chart functions) — Phase 2. Best done after consolidation (Phase 1).

## Problem

No function in \`R/functions/main-functions.R\` has documentation. There are no roxygen2 headers, no parameter descriptions, and no usage examples. Contributors must read the implementation to understand what each function does.

## Proposed solution

Add roxygen2-style documentation to all functions:

\`\`\`r
#' Create an SVG clock face showing elapsed time
#'
#' @param hours Numeric. Hours to display (0-12).
#' @param minutes Numeric. Minutes to display (0-59).
#' @return Character string containing an SVG element.
#' @examples
#' create_clock(1, 30)  # Shows 1:30
create_clock <- function(hours, minutes) { ... }
\`\`\`

## Success criteria

- [ ] All functions have \`@param\`, \`@return\`, and \`@examples\` documentation
- [ ] Documentation accurately describes current behavior
- [ ] Parameter types and expected ranges are specified

## Size: S

Issue #82, opened 2026-03-27
Add CSS custom properties (design tokens) for repeated colour values


lddurbin:
## Parent issue

Extracted from #49 (Remove dead CSS, add design tokens, standardise naming) — Phase 2.

## Problem

Colour values are hardcoded throughout \`main.css\` with no single source of truth:
- \`#3366cc\` appears 5 times (funnel blue)
- \`#cc6633\` appears 2 times (marble brown)
- \`#cc0000\` appears 2 times (outcome red)
- \`#f9f9f9\` appears 2 times (light grey)

Changing a colour requires finding and updating every instance.

## Proposed solution

Add CSS custom properties to \`:root\`:
\`\`\`css
:root {
  --color-funnel: #3366cc;
  --color-marble: #cc6633;
  --color-outcome: #cc0000;
  --color-bg-light: #f9f9f9;
  --color-border-gray: #999;
  --color-target: #90EE90;
}
\`\`\`

Replace all hardcoded instances with \`var(--color-*)\`.

## Success criteria

- [ ] All repeated colour values use CSS custom properties
- [ ] Custom properties defined in \`:root\` block at top of \`main.css\`
- [ ] No visual changes in rendered output
- [ ] Inline colour values in OJS/JS files also updated where practical

## Size: S

Issue #79, opened 2026-03-27
Tag each deployment in git for traceability


lddurbin:
## Parent issue

Extracted from #45 (Add rollback capability) — Phase 2.

## Problem

There is no way to see which commits were actually deployed to production. Re-deploying a previous version requires finding the right commit manually.

## Proposed solution

Add a step to \`deploy.yml\` after successful rsync that creates and pushes a git tag:

\`\`\`yaml
- name: Tag deployment
  run: git tag "deploy-\$(date +%Y%m%d%H%M%S)" && git push --tags
\`\`\`

## Success criteria

- [ ] Each successful deployment creates a \`deploy-YYYYMMDDHHMMSS\` tag
- [ ] Tags are pushed to the remote repository
- [ ] Failed deployments do not create tags

## Size: XS

Issue #69, opened 2026-03-27
Audit scripts/ directory and archive obsolete conversion tools


lddurbin:
## Problem

The `scripts/` directory has accumulated tools from multiple conversion approaches during the project's evolution. Files like `extract-content-simple.py`, `extract-content.py`, `smart-extract.py`, and `streamlined-converter.py` suggest iterative experimentation — some are likely obsolete, superseded by the current workflow documented in `workflow/CONVERSION_GUIDE.md`.

Keeping dead scripts in the active directory makes it unclear which tools are current and supported.

## Proposed solution

1. Review each script in `scripts/` against the current conversion workflow
2. Identify which scripts are actively used or referenced
3. Move obsolete scripts to `scripts/archive/` with a brief README noting why they were archived
4. Keep only current, actively-used scripts in the main `scripts/` directory

## Success criteria

- [ ] Every script in `scripts/` is either actively used or moved to `scripts/archive/`
- [ ] `scripts/archive/README.md` explains what the archived scripts were for
- [ ] No references to archived scripts remain in CLAUDE.md or workflow docs
- [ ] Conversion workflow still functions correctly after cleanup

## Size: S

Requires reviewing ~20 scripts, checking usage, and moving files. Low-moderate token usage — mostly reading and file operations.

PR #106, opened 2026-04-02
Improve maintainability: CI, CSS tokens, scripts, R docs

### LINKED ISSUE #54: Remove or gate daily scheduled build trigger
### Description

The deploy workflow runs daily at 11:00 UTC (`deploy.yml:9`):

```yaml
schedule:
  - cron: '0 11 * * *'
```

This triggers a full build and deploy even when no code has changed, wasting CI minutes and creating unnecessary deployment risk (particularly if a floating dependency like R `'release'` changes unexpectedly).

### Affected files

- `.github/workflows/deploy.yml:8–9`

### Recommended fix

Either remove the schedule trigger entirely (deploys on push to main are sufficient) or add a change-detection gate:

```yaml
- name: Check for changes since last deploy
  id: check
  run: |
    LAST_TAG=$(git describe --tags --match "deploy-*" --abbrev=0 2>/dev/null || echo "")
    if [ -n "$LAST_TAG" ] && git diff --quiet "$LAST_TAG"..HEAD; then
      echo "skip=true" >> $GITHUB_OUTPUT
    fi
```

### Success criteria

- [ ] No unnecessary deployments when code hasn't changed
- [ ] CI minutes not wasted on daily no-op builds

lddurbin:
## Summary
- **#54**: Remove daily cron schedule from deploy workflow — push-to-main triggers are sufficient, saving CI minutes
- **#79**: Tag each successful deployment with `deploy-YYYYMMDDHHMMSS` git tags for traceability and rollback
- **#82**: Add CSS custom properties (`:root` design tokens) for 6 repeated colour values, replacing hardcoded hex codes
- **#56**: Remove `!important` from `.fe-row-outcome` by scoping selector under `.fe-data-table` for natural specificity
- **#69**: Archive 10 obsolete conversion scripts to `scripts/archive/` with explanatory README
- **#85**: Add roxygen2 documentation (`@param`, `@return`, `@examples`) to all 7 R helper functions

Closes #54, #56, #69, #79, #82, #85

## Test plan
- [x] All 40 JS tests pass
- [x] All 33 R tests pass
- [ ] Verify funnel experiment outcome row still has yellow background (no `!important` regression)
- [ ] Verify CSS custom properties render correctly across browsers
- [ ] Confirm deploy workflow triggers only on push/manual dispatch


lddurbin:
Thanks for the review — all three points were valid. Fixed in a24a01e:

1. **Permissions**: Added `contents: write` to the `build-and-deploy` job so `git push --tags` has the necessary token scope.

2. **Archive README accuracy**: The 8 "missing" files were actually on disk (moved with `mv`) but not tracked because `/scripts/` is in `.gitignore`. Force-added them with `git add -f` so the archive is complete in git.

3. **`--color-target` removed**: It was only used in JS files (`funnel-experiment.js`), not via `var()` in CSS. Removed from `:root` to keep tokens honest. If we later want JS to read CSS tokens, we can re-add it then.

PR #109, opened 2026-04-02
Validate localStorage fields to prevent stored XSS (#16)

### LINKED ISSUE #16: Validate localStorage dice sequence fields to prevent stored XSS
## Description

The `loadDiceSequence()` function in `funnel-experiment.js` validates that localStorage data is an array of the correct length, but does not validate individual fields. The `direction` field (a string like `"3R"` or `"↓"`) is later interpolated via string concatenation into an HTML table and rendered via `innerHTML`. If someone edits localStorage to inject HTML/script content in the `direction` field, it would execute.

**Practical risk is low** (requires the attacker to already have access to the user's browser — this is "self-XSS"), but defense-in-depth suggests closing the gap.

**File:** `assets/scripts/funnel-experiment.js:143` (load) and `:285` (render)

## Fix

Add per-field type validation after the `Array.isArray()` check in `loadDiceSequence()`:

```js
if (!parsed.every(d =>
  typeof d.die1 === 'number' && typeof d.die2 === 'number' &&
  typeof d.diceScore === 'number' && typeof d.displacement === 'number' &&
  typeof d.direction === 'string' && /^(\d+[LR]|↓)$/.test(d.direction)
)) return null;
```

Additionally, consider HTML-escaping string values in `renderDataTable()` before inserting them into the HTML string as a secondary defense layer.

## Success Criteria

- [ ] `loadDiceSequence()` validates all fields of each array element by type
- [ ] The `direction` field is validated against the expected pattern (`/^(\d+[LR]|↓)$/`)
- [ ] Invalid localStorage data is rejected (function returns `null`, triggering fresh generation)
- [ ] The funnel experiment works correctly after the change with valid data
- [ ] Manually setting `direction` to `<script>alert(1)</script>` in localStorage causes the data to be rejected and regenerated

lddurbin:
## Summary
- Adds per-field type validation in `loadDiceSequence()` — rejects localStorage data where any numeric field isn't a number or `direction` doesn't match `/^(\d+[LR]|↓)$/`
- Adds `escapeHTML()` helper and applies it to the `direction` field in `renderDataTable()` as defense-in-depth
- Adds 18 new tests covering validation, rejection of XSS payloads, and HTML escaping

Closes #16

## Test plan
- [x] All 54 tests pass (`npm test`)
- [ ] Verify funnel experiment works normally with valid localStorage data
- [ ] Manually set `direction` to `<script>alert(1)</script>` in localStorage → confirm data is rejected and regenerated

### DIFF
--- a/assets/scripts/funnel-experiment.js
+++ b/assets/scripts/funnel-experiment.js
@@ -142,7 +142,12 @@ export function loadDiceSequence() {
     const stored = localStorage.getItem(STORAGE_KEY);
     if (stored) {
       const parsed = JSON.parse(stored);
-      if (Array.isArray(parsed) && parsed.length === TOTAL_STAGES) {
+      if (Array.isArray(parsed) && parsed.length === TOTAL_STAGES &&
+          parsed.every(d =>
+            typeof d.die1 === 'number' && typeof d.die2 === 'number' &&
+            typeof d.diceScore === 'number' && typeof d.displacement === 'number' &&
+            typeof d.direction === 'string' && /^(\d+[LR]|↓)$/.test(d.direction)
+          )) {
         return parsed;
       }
     }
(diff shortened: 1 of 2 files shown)


lddurbin:
Both points addressed in 175fe5e:

1. **Single-quote escaping** — `escapeHTML` now includes `.replace(/'/g, "&#39;")`, making it safe in attribute contexts too.

2. **`Number.isFinite` guard** — all four numeric fields now use `Number.isFinite()`. As the review notes, `JSON.parse` can't produce `NaN`/`Infinity` (they serialise to `null`), so the `typeof` check already blocks the JSON pathway. But the guard hardens against any future non-JSON source. Added a test confirming that `NaN` (which round-trips to `null` via JSON) is rejected.

PR #110, opened 2026-04-02
Add skip-to-content link for keyboard navigation

### LINKED ISSUE #35: Add skip-to-content link for keyboard navigation
## Severity: Medium (WCAG 2.4.1 — Bypass Blocks)

The Cosmo theme does not provide a skip-to-content link. Keyboard users must tab through the entire sidebar navigation before reaching the main content on every page.

### Recommended fix

Add via `header-includes` in `_quarto.yml`:
```html
<a href="#quarto-document-content" class="skip-link">Skip to content</a>
```

And in `assets/styles/main.css`:
```css
.skip-link {
  position: absolute;
  top: -40px;
  left: 0;
  padding: 8px 16px;
  z-index: 100;
  background: #000;
  color: #fff;
  text-decoration: none;
  font-weight: bold;
}
.skip-link:focus {
  top: 0;
}
```

### Success criteria

- [ ] A "Skip to content" link is the first focusable element on every page
- [ ] The link is visually hidden until focused via keyboard
- [ ] Activating the link moves focus to the main content area
- [ ] The target element (`#quarto-document-content`) exists in rendered HTML

lddurbin:
## Summary
- Adds a skip-to-content link as the first focusable element on every page, satisfying WCAG 2.4.1 (Bypass Blocks)
- The link is visually hidden off-screen and only appears when focused via keyboard Tab
- Targets `#quarto-document-content` which Quarto already renders as the `<main>` element

Closes #35

## Test plan
- [ ] Run `quarto render` and verify the skip link appears in rendered HTML
- [ ] Open any page in browser, press Tab — "Skip to content" link should appear at top-left
- [ ] Press Enter on the link — focus should move to the main content area
- [ ] Verify the link is not visible during normal mouse browsing

### DIFF
--- a/_quarto.yml
+++ b/_quarto.yml
@@ -78,9 +78,9 @@ format:
     theme: cosmo
     css: assets/styles/main.css
     header-includes: |
-      <a href="#quarto-document-content" class="skip-link">Skip to content</a>
       <meta http-equiv="Content-Security-Policy" content="default-src 'self'; script-src 'self' 'unsafe-inline' 'unsafe-eval' https://cdn.jsdelivr.net; style-src 'self' 'unsafe-inline'; font-src 'self'; frame-src https://excalidraw.com; img-src 'self' data:; connect-src 'self' https://cdn.jsdelivr.net;">
       <script src="/assets/scripts/functions.js"></script>
+      <script>document.addEventListener("DOMContentLoaded",function(){var a=document.createElement("a");a.href="#quarto-document-content";a.className="skip-link";a.textContent="Skip to content";document.body.insertBefore(a,document.body.firstChild)});</script>
     number-sections: false
 
 
--- a/_quarto.yml
+++ b/_quarto.yml
@@ -80,7 +80,15 @@ format:
     header-includes: |
       <meta http-equiv="Content-Security-Policy" content="default-src 'self'; script-src 'self' 'unsafe-inline' 'unsafe-eval' https://cdn.jsdelivr.net; style-src 'self' 'unsafe-inline'; font-src 'self'; frame-src https://excalidraw.com; img-src 'self' data:; connect-src 'self' https://cdn.jsdelivr.net;">
       <script src="/assets/scripts/functions.js"></script>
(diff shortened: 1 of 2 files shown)


lddurbin:
Thanks for the thorough review. Here's what I've done:

**1. Skip link placement (critical) — Fixed ✅**

You're right that `header-includes` targets `<head>`. I initially tried `include-before-body` per your suggestion, but Quarto injects that content *inside* `<main>` — after the entire sidebar — which defeats the purpose.

Instead, I'm using a small `DOMContentLoaded` script in `header-includes` that creates the skip link and prepends it as the first child of `<body>` via `insertBefore(a, document.body.firstChild)`. This guarantees valid HTML and correct tab order (before sidebar/navbar).

**2. `tabindex="-1"` on target — Skipping**

The target `#quarto-document-content` renders as `<main>`, which modern browsers handle natively for focus via fragment navigation. Adding `tabindex="-1"` would require either modifying Quarto's template or adding JS to patch the attribute — unnecessary complexity for a `<main>` element. If we see cross-browser issues we can revisit.

**3. `z-index: 100` → `1050` — Fixed ✅**

Good catch. Bootstrap uses `z-index: 1030` for fixed navbars, so `100` would have been hidden behind it. Bumped to `1050`.

lddurbin:
Thanks for the follow-up.

**1. JS-dependency — Acknowledged, no action**

Agreed this is a pragmatic trade-off. The site already requires JS for interactive activities, and the reviewer correctly notes it's not blocking. If we ever move to Quarto template overrides for other reasons, we can add a static fallback then.

**2. Inline script formatting — Fixed ✅**

Expanded to readable multi-line in the YAML block scalar.

**3. Missing `outline` on `:focus` — Fixed ✅**

Added `outline: 3px solid #fff` with `outline-offset: 2px` on `.skip-link:focus`. Bootstrap does have an outline reset, so this is a good defensive measure for an accessibility feature.

lddurbin:
**`position: absolute` → `position: fixed` — Fixed ✅**

Good catch. Bootstrap uses `position: relative` on 35+ selectors, so any positioned ancestor between `<body>` and the skip link would shift the `top: 0` reference point away from the viewport. `position: fixed` eliminates that risk entirely.

PR #111, opened 2026-04-02
Add ARIA roles to custom callout divs

### LINKED ISSUE #36: Add ARIA roles to custom callout divs (.thought, .workbook_callout, etc.)
## Severity: Medium (WCAG 1.3.1 — Info and Relationships)

Custom styled `<div>` elements used throughout the book have no ARIA roles, making their purpose invisible to screen readers. These appear 50+ times across all content.

### Affected classes

| Class | Suggested Fix | Approximate Count |
|---|---|---|
| `.thought` | `role="note"` or `<aside>` | 20+ |
| `.thought_commentary` | `role="note"` | 15+ |
| `.technical_aid` | `role="complementary"` or `<aside>` | 5+ |
| `.workbook_callout` | `role="note"` | 30+ |
| `.neave_note` | `role="note"` | 10+ |

### Recommended fix

Option A — Add `role` attribute:
```html
<div class="thought" role="note" aria-label="Activity">
```

Option B — Use semantic HTML:
```html
<aside class="thought" aria-label="Activity">
```

### Success criteria

- [ ] Every `.thought`, `.thought_commentary`, `.technical_aid`, `.workbook_callout`, and `.neave_note` div has an appropriate ARIA `role`
- [ ] Screen reader announces the purpose of each callout region
- [ ] No duplicate landmark roles on the same page (use `role="note"` rather than `role="complementary"` if there are many)

lddurbin:
## Summary
- Adds `role="note"` and descriptive `aria-label` attributes to all custom callout `<div>` elements across the book (`.thought`, `.thought_commentary`, `.technical_aid`, `.workbook_callout`, `.neave_note`)
- 109 instances updated across 28 files (Days 1–5, index, and chapter template)
- Also normalises previously unquoted `class=foo` attributes to `class="foo"` for consistency

Closes #36

## Details

Each class gets a descriptive `aria-label` so screen readers announce meaningful context:

| Class | `aria-label` |
|---|---|
| `.thought` | "Thought" |
| `.thought_commentary` | "Commentary" |
| `.technical_aid` | "Technical aid" |
| `.workbook_callout` | "Workbook reference" |
| `.neave_note` | "Author's note" |

Uses `role="note"` (non-landmark) rather than `role="complementary"` to avoid polluting landmark navigation on pages with many callouts.

## Test plan
- [x] `quarto render` succeeds with no errors
- [x] Verified 108 `role="note"` attributes appear in rendered HTML across 27 output files
- [x] Verified zero remaining untagged callout divs in source `.qmd` files
- [ ] Screen reader testing: Tab through a content-heavy page (e.g. Day 3 Chapter 5) and verify callouts are announced with their labels

### DIFF
--- a/assets/templates/chapter-template.qmd
+++ b/assets/templates/chapter-template.qmd
@@ -28,7 +28,7 @@ Content goes here when there's no clock.
 :::: {.columns}
 
 ::: {.column width="85%"}
-<div class=thought_commentary>
+<div class="thought_commentary" role="note" aria-label="Commentary">
 
 # PAUSE FOR THOUGHT OR ACTIVITY
 
--- a/content/days/day-01/07-quality-theory.qmd
+++ b/content/days/day-01/07-quality-theory.qmd
@@ -23,9 +23,9 @@ The second mention was in a phrase that I had heard from him time and time again
 
 <span class=deming_quote>“Quality is made at the top.”</span>
 
-<div class=workbook_callout>Pause for Thought 1-a which follows is also (somewhat unsurprisingly!) on [Workbook page 1](workbook.html).</div>
+<div class="workbook_callout" role="note" aria-label="Workbook reference">Pause for Thought 1-a which follows is also (somewhat unsurprisingly!) on [Workbook page 1](workbook.html).</div>
 
-<div class=neave_note>The red box indicates there will now be a Pause for Thought or Activity which is immediately followed by
+<div class="neave_note" role="note" aria-label="Author's note">The red box indicates there will now be a Pause for Thought or Activity which is immediately followed by
 some relevant commentary. My commentary will be in the subsequent shaded red box. So always try to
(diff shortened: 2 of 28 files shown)


lddurbin:
Thanks for the review! Two points addressed:

**Missed `<span class="neave_note">` in day-02/05** — Good catch. Fixed in 9dd3c52. Converted the `<span>` to a `<div>` with `role="note" aria-label="Author's note"`, and replaced the inline `<p>` elements with proper Markdown paragraphs. This fixes both the missing ARIA attributes and the invalid HTML (block-level elements inside an inline element).

**Duplicate `aria-label` values** — Agreed this is a known trade-off and acceptable for this first pass. Unique labels per callout (e.g. `aria-label="Workbook reference: Pause for Thought 2-a"`) would be a nice improvement but would require hand-editing ~50 instances with context-specific text. I'll note this as a follow-up on the issue.

PR #115, opened 2026-04-03
Remove unnecessary renv sandbox disable from CI

### LINKED ISSUE #53: Investigate and fix renv sandbox issue instead of disabling it
### Description

The CI pipeline disables the renv sandbox (`deploy.yml:75`):

```bash
echo 'Sys.setenv(RENV_CONFIG_SANDBOX_ENABLED = "FALSE")' >> .Rprofile
```

The sandbox exists to prevent system R packages from leaking into the project environment. Disabling it reduces package isolation safety and means the CI environment may behave differently from local development.

### Affected files

- `.github/workflows/deploy.yml:75`

### Recommended fix

Investigate why the sandbox causes issues (likely related to igraph compilation — see also M6/igraph workaround). Fix the underlying cause rather than disabling the safety mechanism.

### Success criteria

- [ ] `RENV_CONFIG_SANDBOX_ENABLED` is not set to FALSE in deployment
- [ ] `renv::restore()` succeeds with sandbox enabled
- [ ] Build produces same results with sandbox on/off

lddurbin:
## Summary

- Removes the `RENV_CONFIG_SANDBOX_ENABLED = "FALSE"` workaround from `deploy.yml`
- This was added during early CI debugging ("More debugging" commit) and never cleaned up
- The `deploy-optimized.yml` workflow already runs successfully without disabling the sandbox, confirming it's not needed

Closes #53

## Test plan

- [ ] CI pipeline passes with sandbox enabled (this PR is the test — `renv::restore()` must succeed)
- [ ] Build produces the same output (smoke test step validates this)

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -94,10 +94,6 @@ jobs:
         sudo dpkg -i quarto-1.4.549-linux-amd64.deb
         quarto --version
         
-    - name: Disable renv sandbox
-      run: |
-        echo 'Sys.setenv(RENV_CONFIG_SANDBOX_ENABLED = "FALSE")' >> .Rprofile
-
     - name: Clean renv library
       run: |
         rm -rf renv/library
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -94,10 +94,6 @@ jobs:
         sudo dpkg -i quarto-1.4.549-linux-amd64.deb
         quarto --version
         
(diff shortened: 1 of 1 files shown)


lddurbin:
Addressed the pre-existing observation from the review: removed the `rm -rf renv/library` "Clean renv library" step that was defeating both the cache and the igraph pre-install. With this gone, cached packages are preserved across runs and `renv::restore()` becomes a fast verification rather than a full reinstall.

PR #117, opened 2026-04-04
Add transcription validation script

### LINKED ISSUE #63: Add PDF text extraction script for transcription diff validation
## Problem

Validating that QMD transcriptions match source PDFs currently requires manual page-by-page comparison. With 12 days of content (~300+ pages), this is impractical.

## Proposed Solution

Build a script that:
1. Extracts text from source PDFs using `pdftotext` (poppler-utils)
2. Strips QMD markup (YAML front matter, code fences, OJS blocks, markdown syntax) from `.qmd` files to produce plain text
3. Diffs the two outputs and reports suspicious gaps — missing paragraphs, dropped sentences, transposed words

The script doesn't need a perfect match (formatting differs), but should flag cases where a contiguous block of source text has no close match in the corresponding QMD file.

## Acceptance Criteria

- Script accepts a day number (e.g. `./scripts/validate-transcription.sh 5`) and compares all chapters for that day
- Outputs a human-readable report of potential gaps, not a raw diff
- Can be run locally; CI integration is out of scope for this issue
- Documents which PDF files correspond to which days (letter prefix mapping: E=Day2, F=Day3, etc.)

## Notes

- Source PDFs are in `12-Days-to-Deming/` (gitignored, local only)
- `pdftotext` is available via `brew install poppler` on macOS
- Focus on catching *missing* content, not formatting differences

lddurbin:
## Summary

- Adds `scripts/validate-transcription.sh` that compares source PDF text against QMD transcriptions to flag potentially missing content
- Uses `pdftotext` to extract PDF text, strips QMD markup (frontmatter, code fences, HTML, markdown syntax), then does paragraph-level fingerprint matching
- Handles pdftotext ligature issues (fi/fl/ff glyphs silently dropped) by stripping `f` from both sides during normalisation
- Closes #63

## Key design decisions

- **Paragraph-level matching** rather than line-by-line diff, since PDF and QMD have fundamentally different line breaks
- **Sliding window fingerprints** (5 consecutive significant words) for fuzzy matching that tolerates formatting differences
- **Readability filter** rejects paragraphs where <40% of characters are letters (catches garbled page headers/footers)
- **`-layout` mode** with embedded page number stripping for best text extraction fidelity

## Test results on converted days

| Day | Match Rate |
|-----|-----------|
| 1   | 99%       |
| 2   | 98%       |
| 3   | 98%       |
| 4   | 94%       |
| 5   | 98%       |

## Test plan

- [x] Verify script runs on all 5 converted days with high match rates
- [x] Verify error handling: no args, invalid day number, missing content directory
- [x] Verify `pdftotext` dependency check
- [x] Verify PDF letter-prefix mapping (D=Day1 through O=Day12)

### DIFF
--- a/scripts/validate-transcription.sh
+++ b/scripts/validate-transcription.sh
@@ -1,0 +1,393 @@
+#!/usr/bin/env bash
+#
+# validate-transcription.sh — Compare source PDF text against QMD transcriptions
+#
+# Usage: ./scripts/validate-transcription.sh <day-number>
+#   e.g. ./scripts/validate-transcription.sh 3
+#
+# Requires: pdftotext (brew install poppler)
+#
+# PDF letter-prefix mapping:
+#   D=Day1, E=Day2, F=Day3, G=Day4, H=Day5, I=Day6,
+#   J=Day7, K=Day8, L=Day9, M=Day10, N=Day11, O=Day12
+
+set -euo pipefail
+
+# Force C locale to avoid multibyte issues with pdftotext output
+export LC_ALL=C
+
+# ── Configuration ──────────────────────────────────────────────
+
+REPO_ROOT="$(cd "$(dirname "$0")/.." && pwd)"
+PDF_DIR="$REPO_ROOT/12-Days-to-Deming/PDFs"
+CONTENT_DIR="$REPO_ROOT/content/days"
+
+# Map day numbers (1-12) to PDF letter prefixes (D-O)
+# ASCII: D=68, day 1 → offset 0 → D, day 2 → offset 1 → E, etc.
+day_to_prefix() {
+  local day=$1
+  printf "\\$(printf '%03o' $((67 + day)))"
+}
+
+# Minimum paragraph length (chars) to consider for matching
+MIN_PARA_LEN=40
+
+# ── Helpers ────────────────────────────────────────────────────
+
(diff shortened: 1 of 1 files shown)


lddurbin:
Thanks for the thorough review. Addressed 4 of 7 items:

**Fixed:**
- **PDF extracted twice** — removed unused `pdf_raw.txt` write
- **`tmpdir` not `local`** — added `local` with comment explaining trap visibility
- **Useless `cat`** — replaced with input redirection
- **`ls` parsed for PDF glob** — switched to array glob for `pdf_file`

**Not fixing (with rationale):**
- **Performance (subprocess per paragraph):** The script runs in <5 seconds on the largest day (Day 3, 328 paragraphs). Optimising the hot loop with awk substring matching would sacrifice the `.*` regex flexibility that handles word reordering across PDF/QMD boundaries. Not worth the complexity for a local-only tool.
- **Quarto div content passed through:** The `:::` callout/panel divs contain actual course prose (activities, commentary, notes) that IS transcribed from the PDF. Filtering them out would cause false negatives — flagging content as missing when it's actually present in QMD callouts. The current behaviour is intentional.
- **`qmd_files` line count:** Agreed it's technically fragile but works correctly for all cases here. Not worth restructuring for a display-only count.

PR #120, opened 2026-04-05
Fix Quarto callout-note text overlapping subsequent elements

lddurbin:
## Summary
- The project's `.callout` base CSS rule (`height: 40px`) was colliding with Quarto's native `:::{.callout-note}` blocks, which also receive the `.callout` class in rendered HTML
- This constrained multi-line callout-note content to 40px height, causing text to overflow behind workbook callouts and activity boxes (visible on Day 6 introduction page)
- Added a `.callout.callout-note` override to reset height, line-height, text-align, font-style, and border for Quarto's native callout blocks

## Test plan
- [ ] Run `quarto preview` and check Day 6 introduction page — the callout-note before Activity 6-a should no longer overlap the yellow workbook callout and green activity box
- [ ] Verify other callout-note instances across Day 6 (chapters 2 and 3) render correctly
- [ ] Verify custom callouts (workbook_callout, return_callout, info_callout) still display as single-line yellow/blue/green bars

### DIFF
--- a/assets/styles/main.css
+++ b/assets/styles/main.css
@@ -67,6 +67,15 @@
   background-color: #c2d69b;
 }
 
+/* Override base .callout styles for Quarto's native callout blocks */
+.callout.callout-note {
+  height: auto;
+  line-height: normal;
+  text-align: left;
+  font-style: normal;
+  border: none;
+}
+
 /* ===== FLOATING BOX COMPONENTS ===== */
 /*
 Usage:
--- a/assets/styles/main.css
+++ b/assets/styles/main.css
@@ -68,12 +68,15 @@
 }
 
 /* Override base .callout styles for Quarto's native callout blocks */
-.callout.callout-note {
+.callout.callout-note,
+.callout.callout-warning,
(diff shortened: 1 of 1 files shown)


lddurbin:
Thanks for the review — addressed points 1 and 2:

**Point 1 (border):** Agreed. Removed `border: none` so Quarto's native left-accent border is preserved.

**Point 2 (other callout types):** Agreed. Extended the override to cover all five Quarto callout types (`callout-note`, `callout-warning`, `callout-important`, `callout-tip`, `callout-caution`).

**Point 3 (scope base `.callout`):** Pushing back on this one. The custom callout divs use `class="workbook_callout"` (not `class="callout workbook_callout"`), so `.callout:is(.workbook_callout, ...)` would never match them. The `.callout` base rule is effectively dead code for our custom callouts — it only accidentally hits Quarto's native elements. Scoping it with `:is()` would still leave it as dead code. A separate cleanup to move the base styles into each specific class (and remove the `.callout` rule) could be done later, but it's out of scope for this bug fix.

Issue #24, opened 2026-03-02
Add build verification before rsync --delete


lddurbin:
## Description

The rsync deploy uses `--delete`, meaning a failed build producing an empty or partial `_book/` directory could wipe the production site. The current verification step only checks that the `_book` directory exists, not that it contains meaningful content.

**File:** `.github/workflows/deploy.yml:166`

## Fix

Add a more robust build verification step before the rsync command:
```yaml
- name: Verify build output
  run: |
    HTML_COUNT=$(find _book -name "*.html" | wc -l)
    echo "Found $HTML_COUNT HTML files"
    if [ "$HTML_COUNT" -lt 10 ]; then
      echo "ERROR: Build produced fewer than 10 HTML files — aborting deploy"
      exit 1
    fi
```

## Success Criteria

- [ ] The deploy workflow checks for a minimum number of HTML files before rsync
- [ ] An intentionally broken build (e.g. empty `_book/`) is caught and the deploy is aborted
- [ ] Normal builds still deploy successfully

Issue #23, opened 2026-03-02
Add Dependabot config for GitHub Actions


lddurbin:
## Description

No `.github/dependabot.yml` config exists. GitHub Actions are not automatically checked for known vulnerabilities or available updates.

## Fix

Create `.github/dependabot.yml`:
```yaml
version: 2
updates:
  - package-ecosystem: "github-actions"
    directory: "/"
    schedule:
      interval: "weekly"
```

Note: Dependabot does not natively support `renv` for R packages. R dependency scanning would require a separate solution if desired.

## Success Criteria

- [ ] `.github/dependabot.yml` exists and is valid
- [ ] Dependabot creates PRs for outdated GitHub Actions (verify after a week)
- [ ] The `github-actions` ecosystem is configured with a weekly schedule

Issue #21, opened 2026-03-02
Add venv/ and temp/ to .gitignore


lddurbin:
## Description

`venv/` relies on its own internal `.gitignore` (fragile — if removed, `git add .` could stage Python binaries). `temp/` has analysis artifacts already committed that should not be tracked. Both should be in the root `.gitignore` for defense in depth.

**File:** `.gitignore`

## Fix

Add to `.gitignore`:
```
venv/
temp/
```

Consider also removing the already-tracked `temp/` files from git history if they contain no useful information.

## Success Criteria

- [ ] `.gitignore` contains entries for `venv/` and `temp/`
- [ ] `git status` does not show any untracked files from those directories
- [ ] No accidental commits of Python virtual environment files or temporary analysis artifacts are possible

PR #140, opened 2026-04-16
Harden CI/CD pipeline: permissions, dependabot, build verification

### LINKED ISSUE #20: Add explicit permissions block to deploy workflows
## Description

Neither deploy workflow declares a `permissions:` block, so they run with GitHub's default GITHUB_TOKEN permissions. Best practice is to explicitly declare the minimum permissions needed, reducing blast radius if the token is ever exposed.

**Files:** `.github/workflows/deploy.yml`, `.github/workflows/deploy-optimized.yml`

## Fix

Add to both workflows at the top level:
```yaml
permissions:
  contents: read
```

## Success Criteria

- [ ] Both workflow files contain an explicit `permissions:` block
- [ ] Permissions are set to the minimum required (`contents: read`)
- [ ] Deploy succeeds after the change

lddurbin:
## Summary
- **#20**: Add top-level `permissions: contents: read` to both deploy workflows, limiting default GITHUB_TOKEN scope. The `build-and-deploy` job in `deploy.yml` retains its existing `contents: write` override (needed for `git push --tags`).
- **#21**: Add `venv/` and `temp/` to `.gitignore` for defense in depth (neither directory has tracked files).
- **#23**: Add `.github/dependabot.yml` for weekly GitHub Actions dependency scanning.
- **#24**: Add minimum HTML file count check (≥10) to the smoke test in both workflows, preventing `rsync --delete` from wiping production if a build produces empty/partial output.

Closes #20, #21, #23, #24

## Test plan
- [ ] Verify `deploy.yml` YAML is valid and CI runs pass
- [ ] Verify `deploy-optimized.yml` YAML is valid
- [ ] Confirm Dependabot creates its first scan after merge
- [ ] Confirm `.gitignore` additions don't affect tracked files

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
--- a/.github/workflows/deploy-optimized.yml
+++ b/.github/workflows/deploy-optimized.yml
@@ -119,6 +119,7 @@ jobs:
 
         HTML_COUNT=$(find _book -name "*.html" | wc -l)
         echo "Found $HTML_COUNT HTML files"
+        # Minimum expected HTML files — prevents rsync --delete from wiping production on partial builds
         if [ "$HTML_COUNT" -lt 10 ]; then
           echo "FAIL: Build produced fewer than 10 HTML files — aborting deploy"
           exit 1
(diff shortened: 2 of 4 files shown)


lddurbin:
Added the inline comment on the HTML count threshold as suggested. The other observations (Pages deploy note, find counting subdirs) are noted but no action needed — agree they're non-issues for this PR.

PR #146, opened 2026-04-16
Consolidate two deployment workflows into one

### LINKED ISSUE #59: Resolve two competing deployment workflows
### Description

Two deployment workflows exist:

1. **`deploy.yml`** (178 lines) — Active; triggers on push to main, schedule, and manual dispatch
2. **`deploy-optimized.yml`** (98 lines) — Only triggers on manual dispatch

The optimised workflow has improvements:
- Better caching (`actions/cache@v4` vs v3)
- Does not disable renv sandbox
- Does not pre-install igraph
- Cleaner structure

It's unclear which is canonical, and having two workflows creates confusion about which deploy process is authoritative.

### Affected files

- `.github/workflows/deploy.yml`
- `.github/workflows/deploy-optimized.yml`

### Recommended fix

Decide on one workflow:
- If `deploy-optimized.yml` is the intended successor, promote it (add push/schedule triggers) and remove `deploy.yml`
- If it's experimental, add a comment header explaining its status and when it should be tested

### Success criteria

- [ ] Only one active deployment workflow, or both clearly documented with distinct purposes

lddurbin:
## Summary
- Merges `deploy.yml` and `deploy-optimized.yml` into a single `deploy.yml`, taking the best of each
- Keeps push-to-main trigger and deploy tagging from the original
- Adopts RSPM binary packages, system deps/Quarto/renv caching, and cleaner structure from the optimized version
- Removes 6 debugging/troubleshooting steps that were never cleaned up
- Net reduction: 395 → 136 lines across the two files

Closes #59

## Test plan
- [ ] CI passes on this PR
- [ ] Trigger a manual `workflow_dispatch` after merge to verify end-to-end deploy


lddurbin:
Thanks for the review. Here's how I addressed each point:

**1. `contents: read` vs `git push --tags`** — No change needed. The `build-and-deploy` job already declares `permissions: contents: write` at the job level (line 36 in the original commit), which overrides the workflow-level `contents: read`. The `test-js` job correctly inherits only `read`. This is working as intended.

**2. Apt cache static key** — Agreed, removed entirely in fab34ce. The ~20-30s install time isn't worth the stale-security risk.

**3. Removed HTML-count comment** — Restored in fab34ce. The "prevents rsync --delete from wiping production" rationale is worth keeping since it explains a non-obvious safety invariant.

**4. `repos` arg on `install.packages`** — No change. With `use-public-rspm: true` already set on the R setup step, the RSPM repo is configured as the default. Adding an explicit `repos` is redundant.

PR #177, opened 2026-04-17
Make funnel status updates reliably announce to screen readers (#161)

### LINKED ISSUE #161: Add aria-live regions to funnel experiment status updates
## Problem

The funnel experiment in Day 3 (`assets/scripts/funnel-experiment.js`) updates a status display after each user action ("Drop 1 of 50 complete", "Funnel moved by …"). Without `aria-live`, screen-reader users have to manually re-read the status region to find out what changed — they'll often miss the announcement entirely.

## Scope

- Add `aria-live="polite"` and `aria-atomic="true"` to the `.fe-status` element (`assets/styles/main.css:592-604`).
- Verify the JS in `assets/scripts/funnel-experiment.js` updates the existing element (rather than replacing it), so the live region keeps working.

- **Cost**: ~5 lines, possibly a small refactor of how the script writes status.
- **Benefit**: Screen-reader users following Day 3 hear status updates announced automatically as they interact with the experiment.
- **Visibility to default users**: **None** — `aria-live` has no visual effect.

## Acceptance

- `.fe-status` is announced by VoiceOver / NVDA when its content changes.
- No double-announcements (single live region per status block).
- Manual test passes with at least one screen reader.

## Parent

#158

lddurbin:
## Summary

- Funnel experiment status bars on Day 3 (chapters 11 and 12) had `aria-live="polite"` attrs, but OJS reactivity replaces the `.fe-status` element on each re-render — screen readers only monitor live regions that persist across mutations, so the announcements were being dropped.
- Added a single persistent, visually-hidden `.fe-sr-live` region created at module load and attached inside Quarto's `#quarto-document-content` landmark. A new `announceFunnelStatus()` helper writes plain-text updates into it.
- Added a 200ms write delay so VoiceOver can finish announcing the web-area landmark (triggered by OJS also destroying the focused "Next Stage" button on re-render) before our polite message enters the queue — without the delay, VO collapses both into a single transition and drops ours.
- Removed the now-misleading `role="status" aria-live="polite" aria-atomic="true"` attributes from the visible `.fe-status` div inside `renderStatusHTML` (they implied a working live region that never worked, and keeping both risks double-announcements in browsers that do pick up dynamically-added regions).
- Announcements are skipped while `counter === 0` so the initial page render (where every OJS cell evaluates eagerly) doesn't spam "Rule 1, stage 0 of 40…" / "Rule 2, stage 0 of 40…" before the user has touched anything.

Closes #161.

## Test plan

- [x] JS module parses (`node --check`)
- [x] Manual: verified with VoiceOver on Day 3 ch. 11 — clicking "Next Stage" now produces a spoken status update ("Rule 2, stage 1 of 40. Marble at 33. Funnel now at 27") after the landmark announcement settles.
- [x] Verified via Safari Web Inspector that `.fe-sr-live` is inserted into the DOM at page load (not on first click) and that its `textContent` updates on each interaction.
- [ ] No visual regression to the blue `.fe-status` bar — same copy, same styling.
- [ ] Covers all four rules' status bars (Rules 1, 2, 3, 4) because they all go through `renderStatusHTML`.

## Known follow-up (out of scope)

OJS also replaces the "Next Stage" button on each re-render, so keyboard focus falls to `<body>` and VoiceOver announces the web-area landmark on every click. That's a separate usability bug and warrants its own issue — the fix is either to preserve focus across re-render or to move the button outside the reactive cell.

### DIFF
--- a/assets/scripts/funnel-experiment.js
+++ b/assets/scripts/funnel-experiment.js
@@ -460,7 +460,10 @@ export function renderStatusHTML(rule, description, counter, visible, totalStage
 
   // Announce state changes to screen readers via the persistent live region.
   // Skip counter === 0 so the initial page render doesn't spam announcements
-  // before the user has interacted.
+  // before the user has interacted. Assumes OJS only re-evaluates this cell
+  // when rule2Counter/rule2Visible (etc.) actually change — i.e. on user
+  // clicks. If a future change introduces non-interaction re-runs with the
+  // same counter, those would produce duplicate announcements.
   if (counter > 0) {
     announceFunnelStatus(
       `Rule ${rule}, stage ${counter} of ${totalStages}. Marble at ${lastOutcome}. ${funnelLabel} ${funnelPos}.`
--- a/assets/scripts/funnel-experiment.js
+++ b/assets/scripts/funnel-experiment.js
@@ -397,6 +397,56 @@ export function renderDataTable(rule, stages, tableIndex) {
   return html;
 }
 
+// --- Screen reader live region ---
+// OJS replaces a cell's output element on each re-render, so an aria-live
+// attribute on the visible `.fe-status` isn't reliably announced — screen
+// readers monitor nodes that persist across mutations. A single persistent,
+// visually-hidden live region appended once to the document is the pattern
+// VoiceOver/NVDA pick up consistently.
+
+let feLiveRegion = null;
+
+function ensureLiveRegion() {
+  if (typeof document === "undefined") return null;
+  if (feLiveRegion && feLiveRegion.isConnected) return feLiveRegion;
+  // Prefer attaching inside the Quarto main-content landmark. VoiceOver
+  // is more reliable monitoring live regions that sit inside <main> than
+  // those orphaned at <body> level.
+  const host = document.getElementById("quarto-document-content") || document.body;
+  if (!host) return null;
+  feLiveRegion = document.createElement("div");
+  feLiveRegion.className = "fe-sr-live";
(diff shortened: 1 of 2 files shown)


lddurbin:
Thanks for the review. Acting on #3; pushing back on #1 and #2.

**#3 Side-effect assumption — taken.** Added a comment above the `announceFunnelStatus` call in `renderStatusHTML` pinning down the assumption that OJS only re-evaluates this cell on user-driven counter changes, and flagging that any future non-interaction re-runs with the same counter would produce duplicate announcements. See b2e82cd.

**#1 `clip` vs `clip-path` — declining.** The existing `.dyslexia-toggle-label` utility in `main.css:816` uses the same `clip: rect(0, 0, 0, 0)` pattern. Migrating only the new `.fe-sr-live` rule would create inconsistency inside one file for no functional gain (the reviewer's own note: "low priority — not broken"). A project-wide refactor to `clip-path: inset(50%)` is a reasonable cleanup but belongs in its own PR so the diff reads as "unified visually-hidden utility" rather than leaking into an a11y fix.

**#2 Magic number `200` → named constant — declining.** The constant would be used in exactly one place, one line below the comment that already explains the *why* ("VoiceOver needs time to finish the landmark announcement triggered by the focused button being replaced"). A name like `VOICEOVER_LANDMARK_SETTLE_MS` restates the comment but doesn't encode anything the comment doesn't already. Extracting also implies reuse that doesn't exist. If AT testing later shows the value needs tuning, renaming one literal is no harder than renaming one constant.

PR #180, opened 2026-04-18
Restore keyboard focus to funnel stage buttons after OJS re-render

### LINKED ISSUE #178: Funnel experiment buttons lose keyboard focus on each click due to OJS re-render
## Problem

In the Day 3 Funnel Experiment (chapters 11 and 12), every click of "Next Stage →" or "Complete Remaining Stages" causes the button itself to be destroyed and recreated. Keyboard focus falls back to `<body>`, and VoiceOver/NVDA announce the web-area landmark ("The first two rules of the funnel — 12 days to deming, web content") on every click.

**Root cause:** the OJS cell that renders the buttons depends on `ruleNCounter`, so every counter change replaces the cell output — including the focused button. Example from `content/days/day-03/11-the-first-two-rules-of-the-funnel.qmd:99-110`:

```js
{
  const el = html`${renderStageButtonsHTML(rule2Counter, TOTAL_STAGES)}`;
  el.querySelector('.fe-stage-next').onclick = () => {
    if (rule2Counter < TOTAL_STAGES) mutable rule2Counter = rule2Counter + 1;
  };
  ...
  return el;  // fresh DOM element on every re-run — old button + focus gone
}
```

Happens for all four rules (Rule 1, Rule 2 in ch. 11; Rule 3, Rule 4 in ch. 12) and uses the shared helper `renderStageButtonsHTML` in `assets/scripts/funnel-experiment.js:431-435`.

## Impact

- **Keyboard users** lose their place in the tab order on every interaction — they can't rapidly step through stages with Space/Enter.
- **Screen-reader users** hear the landmark name re-announced on every click instead of a useful state update (also makes #161's live-region announcement harder to hear since it arrives *after* the landmark chatter).
- **WCAG 2.4.3 (Focus Order)** — focus is moved unexpectedly by a content change.
- **WCAG 3.2.1 (On Focus)** — activating a control causes a context change the user didn't request.

## Fix Options

1. **Re-focus after re-render** (smallest change). Track whether `.fe-stage-next` was focused before the click, then re-focus the new button after OJS inserts it. Could be done via a MutationObserver or a `queueMicrotask` at the end of the click handler.

2. **Update `disabled` in place rather than re-rendering** (structurally cleaner). Render the button structure once in a cell that doesn't depend on `counter`, and have a separate reactive effect that flips the `disabled` attribute when `counter >= totalStages`. The focused DOM node persists across state changes.

3. **Move the click handler outside the reactive cell** and use event delegation on a stable parent — similar in spirit to option 2.

Option 2 or 3 is the right long-term fix. Option 1 is a quick patch if option 2/3 turns out to need a broader OJS restructure.

## Acceptance

- After clicking "Next Stage", keyboard focus remains on the "Next Stage" button (or moves gracefully to "Complete Remaining Stages" when the former becomes disabled on stage 40).
- VoiceOver no longer re-announces the web-area landmark on every click.
- Manual test passes for all four rules on Day 3 ch. 11 and ch. 12.
- No regression in the existing click behaviour (counter increments, cells re-render, status/track/tables update).

## Parent

#158

## Related

Surfaced while fixing #161 (aria-live announcements). The aria-live fix works correctly but is partially drowned out by the landmark chatter this bug causes.

lddurbin:
## Summary

- Adds `restoreStageFocus(currentEl, preferred)` to `assets/scripts/funnel-experiment.js`. The helper captures the cell's output slot at click time (before OJS orphans the old element) and uses a `MutationObserver` to refocus the matching button in the replacement DOM — with a graceful fallback to the sibling button when the preferred one is disabled.
- Wires the helper into the four stage-button OJS cells in Day 3 chapters 11 and 12 (Rules 1–4), so Space/Enter activation no longer drops focus to `<body>`.
- Eliminates the "12 days to deming, web content" landmark re-announcement that VoiceOver emitted on every click — which was both annoying on its own and partially drowned out the aria-live status messages introduced by #161.

## Why

Issue #178: OJS cells re-evaluate whenever their reactive dependencies change. The stage-button cell reads `ruleNCounter`, so mutating the counter replaces the cell's output DOM — including the focused button. Keyboard users lost their place in the tab order on every interaction (WCAG 2.4.3, 3.2.1).

This implements Option 1 from the issue (smallest change). The structurally cleaner Option 2 — flipping `disabled` in place on a persistent DOM node — would require restructuring the OJS cells to break the counter dependency, and is a fine follow-up if we want to silence the per-click "Next Stage, button" role re-announcement VoiceOver still emits.

Closes #178.

## Test plan

- [ ] `quarto preview` the Day 3 chapter 11 page.
- [ ] **Keyboard-only:** Tab to "Next Stage" on the Rule 2 funnel, press Space repeatedly. Confirm each press advances the stage without re-tabbing (focus remains on the button).
- [ ] Repeat for Rule 1 (same page), Rule 3, and Rule 4 (chapter 12).
- [ ] **VoiceOver (Safari):** With VO on, Tab to "Next Stage", activate. Confirm VO no longer announces "12 days to deming, web content" on each click. The expected sequence is "Next Stage, button" → the polite live-region status update.
- [ ] **Complete Remaining Stages:** Click it mid-run. Confirm the button sequence still behaves (counter jumps to 40, both buttons disable; focus lands somewhere reasonable — for now, acceptably lost on this final transition since both are disabled).
- [ ] **Mouse:** Normal click behaviour unchanged; counter still increments, tables/track/status update as before.

### DIFF
--- a/assets/scripts/funnel-experiment.js
+++ b/assets/scripts/funnel-experiment.js
@@ -502,6 +502,12 @@ export function renderStageButtonsHTML(counter, totalStages) {
 // destroyed and focus falls to <body>. Screen readers then announce the
 // web-area landmark on every click (see issue #178).
 //
+// Call this synchronously *after* the `mutable counter = ...` assignment that
+// triggers the re-render. OJS schedules cell re-evaluation on a microtask,
+// so the observer registered here is guaranteed to be in place before the
+// DOM swap. Moving the call before the mutation, or deferring it to a
+// setTimeout, would break that ordering guarantee.
+//
 // currentEl: the button container from the current render. Must be passed in
 //   while it is still mounted, so we can capture `.parentElement` — the
 //   cell's output slot, which persists across re-renders.
@@ -516,7 +522,11 @@ export function restoreStageFocus(currentEl, preferred = ".fe-stage-next") {
   const observer = new MutationObserver(() => {
     const next = cellContainer.querySelector(".fe-stage-next");
     const complete = cellContainer.querySelector(".fe-stage-complete");
-    if (!next && !complete) return;
+    // Wait until both buttons are mounted. `renderStageButtonsHTML` emits
+    // them in a single HTML string so they typically arrive together, but
+    // observer callbacks can fire mid-insertion — guarding on both avoids
+    // focusing a fallback that simply hasn't been mounted yet.
+    if (!next || !complete) return;
 
     observer.disconnect();
     const primary = preferred === ".fe-stage-complete" ? complete : next;
--- a/assets/scripts/funnel-experiment.js
+++ b/assets/scripts/funnel-experiment.js
@@ -496,6 +496,47 @@ export function renderStageButtonsHTML(counter, totalStages) {
        + `<button class="fe-button fe-button-complete fe-stage-complete"${d}>Complete Remaining Stages</button></div>`;
 }
 
+// Restore keyboard focus after an OJS re-render replaces the button container.
+// OJS mutates the counter inside the click handler, which re-evaluates the
+// cell and swaps the returned element — so the just-clicked button is
+// destroyed and focus falls to <body>. Screen readers then announce the
+// web-area landmark on every click (see issue #178).
+//
+// currentEl: the button container from the current render. Must be passed in
+//   while it is still mounted, so we can capture `.parentElement` — the
+//   cell's output slot, which persists across re-renders.
+// preferred: '.fe-stage-next' or '.fe-stage-complete' — the button the user
+//   activated. If it is disabled in the replacement DOM, we fall back to the
+//   other button so keyboard flow continues gracefully.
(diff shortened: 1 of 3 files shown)


lddurbin:
Thanks for the review. Addressed both suggestions in `e2e1071`:

1. **Guard tightened** to `!next || !complete` so the observer waits for both buttons to mount. `renderStageButtonsHTML` emits them in one HTML string, but observer callbacks can fire mid-subtree-insertion, so waiting for both is the correct defensive posture.

2. **Timing assumption documented** in the helper's doc block (opted to put it there rather than at each of the 8 call sites — it's a property of `restoreStageFocus`'s contract, and duplicating across rules adds noise). The note flags that the call must remain synchronous after the mutation, and names the two refactors that would break the guarantee (calling before the mutation, or deferring via setTimeout).

No change on the final-stage focus — the existing comment already flags the histogram-focus follow-up.

PR #181, opened 2026-04-18
Consolidate visually-hidden CSS and migrate off deprecated clip

### LINKED ISSUE #179: Migrate visually-hidden utility from deprecated `clip` to `clip-path: inset(50%)`
## Problem

The project has two copies of the visually-hidden CSS pattern that use the deprecated `clip` property:

```css
clip: rect(0, 0, 0, 0);
```

- `assets/styles/main.css:687` — `.fe-sr-live` (funnel experiment live region)
- `assets/styles/main.css:831` — `.dyslexia-toggle-label` (dyslexia toggle mobile icon-only mode)

The `clip` property is deprecated in favour of `clip-path`. It still works universally today, so this isn't urgent — but while we're at it, unifying on a single utility class would also cut the duplication.

## Scope

1. Consolidate both occurrences into a single reusable `.visually-hidden` (or `.sr-only`) utility class, following the modern pattern:

   ```css
   .visually-hidden {
     position: absolute;
     width: 1px;
     height: 1px;
     padding: 0;
     margin: -1px;
     overflow: hidden;
     clip-path: inset(50%);
     white-space: nowrap;
     border: 0;
   }
   ```

2. Update the two existing sites to use the utility class (keep the existing class names as aliases if they also carry non-hiding styles, or replace them if not).

3. Update `assets/scripts/font-toggle.js` and `assets/scripts/funnel-experiment.js` if the class name changes.

- **Cost**: ~10 lines touched across 1 CSS file and possibly 2 JS files.
- **Benefit**: One utility class instead of two copies; future-proof against `clip` removal; easier for future features that need the same pattern.
- **Visibility to default users**: **None** — purely internal CSS refactor.

## Acceptance

- One canonical visually-hidden utility exists in `main.css`.
- No lingering `clip: rect(...)` declarations.
- Dyslexia toggle still collapses to icon-only on narrow viewports with the label still announced by screen readers.
- Funnel experiment status announcements still fire on Day 3 ch. 11 and 12.

## Context

Raised in the review of #177 (fix for #161). Declined there to keep the PR scoped to the a11y fix; filing as its own small cleanup.

## Parent

#158

lddurbin:
## Summary

- Adds a single canonical `.visually-hidden` utility class to `main.css`, using `clip-path: inset(50%)` instead of the deprecated `clip: rect(...)`.
- Removes the two duplicated hiding rules that previously existed on `.fe-sr-live` and the narrow-viewport `.dyslexia-toggle-label`.
- Wires the utility at its two application sites:
  - `funnel-experiment.js` now adds `visually-hidden` alongside `fe-sr-live` when creating the live region.
  - `font-toggle.js` uses `matchMedia` to toggle `visually-hidden` on the dyslexia label at the 700px breakpoint. Plain CSS lacks media-query-scoped class composition, so the conditional case lives in JS — the rule itself stays in CSS, applied once.

## Why

Issue #179: the project had two copies of the visually-hidden pattern, both using the deprecated `clip` property. Unifying them into one utility cuts the duplication and future-proofs against `clip` removal from browsers.

No user-visible change — purely an internal refactor.

Closes #179.

## Test plan

- [ ] `grep -rn "clip: rect" assets/` returns no matches.
- [ ] `quarto preview` Day 3 chapter 11 (or any page) and confirm:
  - [ ] **Wide viewport (>700px):** the Dyslexia font toggle shows "Aa Dyslexia font" in full.
  - [ ] **Narrow viewport (≤700px):** resize the window below 700px — the label collapses to icon-only, but the button's accessible name still includes "Dyslexia font" (inspect via DevTools or VoiceOver).
  - [ ] **Live resize across the breakpoint:** label hides and shows correctly without a page refresh (confirms `matchMedia` listener is wired).
  - [ ] **Funnel experiment (Day 3 ch. 11 or 12):** click "Next Stage" — the aria-live announcement still fires (confirms the live region is still hidden off-screen but accessible to screen readers).

### DIFF
--- a/assets/scripts/font-toggle.js
+++ b/assets/scripts/font-toggle.js
@@ -46,7 +46,29 @@
     });
 
     document.body.appendChild(button);
+    syncLabelVisibility(button);
     return button;
+  }
+
+  function syncLabelVisibility(button) {
+    // On narrow viewports the button collapses to icon-only, with the
+    // label visually hidden but still announced by screen readers
+    // (aria-label on the button + the live label text both contribute).
+    // matchMedia keeps this in sync across viewport-size changes without
+    // duplicating the hiding rule in CSS.
+    var label = button.querySelector(".dyslexia-toggle-label");
+    if (!label || !window.matchMedia) return;
+    var mq = window.matchMedia("(max-width: 700px)");
+    function sync() {
+      label.classList.toggle("visually-hidden", mq.matches);
+    }
+    sync();
+    if (mq.addEventListener) {
+      mq.addEventListener("change", sync);
+    } else if (mq.addListener) {
+      // Safari < 14 fallback
+      mq.addListener(sync);
+    }
   }
 
   function yieldToPageNav(button) {
--- a/assets/scripts/font-toggle.js
+++ b/assets/scripts/font-toggle.js
@@ -58,6 +58,8 @@
     // duplicating the hiding rule in CSS.
     var label = button.querySelector(".dyslexia-toggle-label");
     if (!label || !window.matchMedia) return;
+    // Breakpoint matches the .dyslexia-toggle @media rule in main.css.
+    // Keep both values in sync if the responsive threshold changes.
     var mq = window.matchMedia("(max-width: 700px)");
     function sync() {
       label.classList.toggle("visually-hidden", mq.matches);
--- a/assets/scripts/funnel-experiment.js
+++ b/assets/scripts/funnel-experiment.js
@@ -415,7 +415,10 @@ function ensureLiveRegion() {
   const host = document.getElementById("quarto-document-content") || document.body;
   if (!host) return null;
   feLiveRegion = document.createElement("div");
-  feLiveRegion.className = "fe-sr-live";
+  // fe-sr-live is the semantic identifier (discoverable in DevTools /
+  // future diagnostic styles); visually-hidden provides the off-screen
+  // positioning defined once in main.css.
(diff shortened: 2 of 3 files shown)


lddurbin:
Thanks — both points were valid and I verified them against the built Bootstrap bundle (`_book/site_libs/bootstrap/bootstrap-*.min.css`). A few things worth noting about what I found:

**Bootstrap's `.visually-hidden` still uses `clip: rect(0, 0, 0, 0) !important`**, not `clip-path`. So our local rule's `clip-path: inset(50%)` genuinely contributes — it's the only place that declaration exists in our cascade — even though the shared properties (width/height/padding/etc.) are overridden by Bootstrap's `!important`.

That means:
- The deprecated `clip` property is still present in the built output, but it's coming from Bootstrap, not our authored CSS. We've met the spirit of the issue (no `clip: rect(...)` in our code) even if we can't fully eliminate it from the bundle.
- Our rule does add real value via `clip-path` — not purely documentary.
- Future divergence on shared properties would need `!important` to win, which I've now flagged in the comment.

Addressed in `73decb4`:
1. Expanded `.visually-hidden` doc block explaining the Bootstrap overlap, what our rule actually contributes, and the `!important` implication for future divergence.
2. Added a sync-comment on the matchMedia call in `font-toggle.js` pointing to the CSS `@media` rule.

Declined to remove the local rule entirely because the issue's acceptance criterion specifically calls for "one canonical visually-hidden utility in `main.css`" — and because our `clip-path` declaration is load-bearing (Bootstrap doesn't provide it).

Issue #182, opened 2026-04-18
Fix AA contrast failures on brand colors (buttons, marble highlights)


lddurbin:
## Problem

Three brand-colored elements fail WCAG AA contrast (< 4.5:1) and are currently baseline-ignored in the pa11y-ci config:

| Selector | Role | Current contrast | Target |
|---|---|---|---|
| \`.btn-primary\` | Bootstrap/Cosmo primary button (e.g. "Download Your Notes") | 3.97:1 (white on \`#2780e3\`) | 4.5:1 |
| \`.fe-marble\` | Orange highlight for "marble" references in the funnel experiment | 3.81:1 (\`#cc6633\` on white) | 4.5:1 |
| \`.fe-button-complete\` | Green "Complete Remaining Stages" button in funnel experiment | 3.13:1 (white on \`#28a745\`) | 4.5:1 |

All three were flagged by \`pa11y-ci\` (WCAG2AA.Principle1.Guideline1_4.1_4_3.G18.Fail) when CI was added in #164.

## Scope

- Darken each color just enough to reach AA, preserving visual identity.
- Candidate targets:
  - \`--color-marble\`: \`#cc6633\` → \`#a64c1f\` (≈5.0:1)
  - \`.fe-button-complete\`: \`#28a745\` → \`#1e7e34\` (≈5.0:1, matches current hover state)
  - \`.btn-primary\`: requires overriding Cosmo's \`--bs-primary\` or darkening the button variant specifically
- Remove the matching \`hideElements\` suppressions from \`.pa11yci.json\` once fixed.

## Parent

#158

## Source

Surfaced by baseline pa11y-ci run in #164.

Issue #159, opened 2026-04-17
Add long-descriptions to charts in Days 2-3 (pattern + first application)


lddurbin:
## Problem

Many control charts, run charts, and process diagrams in the 12 Days carry pedagogy — the learner is supposed to *see* a pattern, a special-cause point, a rule violation. Alt text (currently provided via the `{.lightbox}` pattern in `CONVERSION_GUIDE.md`) is too short to describe what to see. Screen-reader users and learners who can't parse the visual lose the lesson.

## Scope

- Define the long-description pattern: `![alt](img){.lightbox}` followed immediately by `<details><summary>Describe this chart</summary>...</details>`.
- Document the convention in `workflow/CONVERSION_GUIDE.md` so future days follow it.
- Apply it to all chart and diagram images in **Day 2** (red beads) and **Day 3** (funnel) — the two days where charts carry the heaviest pedagogical load.

## Cost / Benefit / Visibility

- **Cost**: Editorial — each chart needs a thoughtful 1-3 sentence description by someone who knows what the learner should see. Roughly 15-30 charts in Days 2-3.
- **Benefit**: Screen-reader users get the chart's pedagogical meaning, not just "Figure 3.2". Cognitively-fatigued readers and chart-novices get a written summary.
- **Visibility to default users**: **None** — `<details>` is collapsed by default; only the small "Describe this chart" disclosure appears under each image.

## Acceptance

- Pattern documented in `CONVERSION_GUIDE.md`.
- Every chart/diagram image in Day 2 and Day 3 has a `<details>` long-description directly after it.
- Decorative images (e.g. portraits, icons) are skipped.

## Parent

#158

Issue #158, opened 2026-04-17
Accessibility roadmap (tracking issue)


lddurbin:
## Background

Following the OpenDyslexic font request (#155), we audited what other accessibility improvements would benefit readers without degrading the default experience. This is the tracking issue for that roadmap.

## Status (2026-04-27)

**Tier 1 + Tier 2 are complete.** Only #170 (TTS, Tier 3) remains, and it is explicitly deferred until reader feedback asks for it.

## Existing accessibility foundations

The site already has:

- Skip-to-content link (`assets/styles/main.css:17-33`)
- Focus indicators on form elements (`main.css:435-443`)
- `prefers-contrast: high` and `prefers-reduced-motion: reduce` support (`main.css:446-473`)
- ≥44×44px mobile touch targets (`main.css:307-310`)
- Self-hosted fonts under a strict CSP

## Roadmap

Issues are grouped by **visibility to default users** — the higher the tier, the lower the visual cost.

### Tier 1 — Invisible to default users (highest leverage, lowest cost)

- [x] #155 — OpenDyslexic font toggle (parent)
- [x] #159 — Long-descriptions for Day 2-3 charts (pattern + first application)
- [x] #160 — Long-descriptions for remaining chart-heavy days
- [x] #161 — `aria-live` regions for funnel experiment
- [x] #162 — `lang` attributes for non-English quotes
- [x] #163 — Redundant non-color cues for callouts
- [x] #164 — Automated a11y checks (pa11y / axe-core) in CI

### Tier 2 — Small visible additions (moderate cost, broad benefit)

- [x] #165 — Light/dark mode toggle
- [x] #166 — Reading-preferences panel (groups font + future toggles)
- [x] #167 — `<dfn>` tooltip mechanism for glossary (later replaced by appendix glossary in #265)
- [x] #168 — Mark up Deming jargon across all days

### Tier 3 — Real UX trade-offs (consider carefully)

- [x] #169 — Estimated reading time per chapter
- [ ] #170 — Per-chapter text-to-speech button (deferred — see issue for rationale)

## Recommended sequence (historical)

The original sequencing — land #147 first, then Tier 1 in parallel, then settle the reading-preferences panel (#166) before the OpenDyslexic toggle (#157) — has all played out. #147 and #157 both closed 2026-04-17; #166 shipped before #157, scoping the toggle as documented.

## Source

Discussion arising from #155 (OpenDyslexic request from alumni Slack), 2026-04-17.

[...]

PR #183, opened 2026-04-18
Add pa11y-ci CI + document lang-attr convention (#164, #162)

### LINKED ISSUE #162: Mark non-English quotes with lang attributes
## Problem

Deming references occasionally include non-English terms — Japanese kaizen vocabulary, German management terms, French philosophical references. Without `lang` attributes, screen readers mispronounce them, which is jarring at best and unintelligible at worst.

## Scope

- Audit all `.qmd` files for non-English words and phrases.
- Wrap each in `<span lang="xx">…</span>` with the appropriate ISO language code.
- Document the convention in `workflow/CONVERSION_GUIDE.md` so future content follows it.

- **Cost**: Audit pass across 80+ qmd files. The actual edits are small.
- **Benefit**: Screen-reader users hear correct pronunciation. SEO and translation tools also benefit.
- **Visibility to default users**: **None** — `lang` has no visual effect.

## Acceptance

- All non-English terms in the 12 Days content are wrapped in `<span lang>` (or have `lang` set on a block element where appropriate).
- Convention documented in `CONVERSION_GUIDE.md`.

## Parent

#158

### LINKED ISSUE #164: Add automated accessibility checks (pa11y or axe-core) to CI
## Problem

There's no automated a11y regression check. Issues like #147 (293 missing input labels) and the heading hierarchy bugs fixed in #149 / #150 could have been caught at build time rather than in evaluation passes.

## Scope

- Choose a tool: [pa11y-ci](https://github.com/pa11y/pa11y-ci) (recommended for static-site simplicity) or [axe-core](https://github.com/dequelabs/axe-core).
- Add a step to `.github/workflows/deploy.yml` (or a separate workflow) that runs the checker against the rendered `_book/` directory.
- Configure rules to match WCAG 2.1 AA.
- Triage the initial baseline: fix what's quick, allow-list what isn't (with linked follow-up issues for each suppression).
- Block PR merges on **regressions**, not on existing baseline issues.

- **Cost**: Setup ~half a day. Initial triage may surface 10-50 findings to categorise. Ongoing: occasional maintenance when rules change.
- **Benefit**: Catches a11y regressions before they ship. Provides a measurable baseline. Validates the work in the rest of this roadmap.
- **Visibility to default users**: **None** — CI-only.

## Acceptance

- CI runs a11y checks on every PR and on `main`.
- A baseline is documented (allowed existing failures, with issue links).
- Build fails on new violations.

## Parent

#158

lddurbin:
## Summary

Bundles two Tier 1 items from the accessibility roadmap (#158): automated a11y checks in CI (#164) and the language-attribute convention (#162).

### #164 — pa11y-ci accessibility checks in CI

Adds `.github/workflows/accessibility.yml` that builds the book and runs [pa11y-ci](https://github.com/pa11y/pa11y-ci) against a representative 22-URL sample on every PR and push to main. Config is at `.pa11yci.json`, WCAG 2.1 AA.

**Baseline triage during setup:**
- **Fixed — Cosmo gray-text contrast (4.39:1 → 7.9:1):** Cosmo's default `.blockquote`, `caption`, `figcaption`, `aside`, `.margin-caption`, `.column-margin`, `.figure-caption` all ship fractionally below AA. Overridden to `var(--bs-gray-700)` in `main.css`. Visual change is minimal — these elements stay distinguishably lighter than body text.
- **Fixed — 7 duplicate-id violations:** `## Heading []{#sec-pageN}` was emitting an empty `<span>` carried into the TOC alongside the heading, producing two elements with the same id. Changed to `## Heading {#sec-pageN}` (heading attribute) across `content/appendix/{03,05,06}-*.qmd`. IDs now live on `<section>`; no TOC duplication.
- **Globally ignored (false positives / scoped elsewhere):**
  - `H91.A.EmptyNoId` — pa11y/HTMLCS doesn't recognise `aria-label` on anchors; fires on Quarto's sidebar accordion toggles which are correctly labelled.
  - `H37` — R ggplot figures missing alt; tracked by #159/#160.
  - `H32.2` — OJS interactive inputs aren't submit-based forms.
- **Baseline-suppressed via `hideElements` with follow-up issue:** `.btn-primary`, `.fe-marble`, `.fe-button-complete` fail AA contrast. Tracked in #182.

**Result:** 22/22 URLs now pass pa11y-ci. Build will fail on any new violation.

### #162 — Audit complete, convention documented

Audit across all `.qmd` files found zero Japanese/German/French/Spanish content. The only Latin was 139× `i.e.` and 61× `e.g.`, which qualify for the WCAG 3.1.2 vernacular exception (and forcing `lang=\"la\"` on them can actually regress screen-reader pronunciation). So no content edits were needed — closing as \"audit-only.\"

Added an \"Accessibility Conventions\" section to `workflow/CONVERSION_GUIDE.md` describing when to use `<span lang=\"xx\">` (e.g. genuine Japanese terms like *kaizen* if they appear in appendices Q–S) and when not to (proper names, vernacular loanwords). Sets the convention before appendix conversion kicks off.

## Test plan

- [ ] CI `Accessibility` workflow runs and passes on this PR
- [ ] Existing `test-js` and `Claude Code Review` workflows continue to pass
- [ ] Spot-check a rendered page (e.g. `content/days/day-01/02-rediscovered.html`) — blockquotes, figcaptions should look visibly darker (but still distinguishably lighter than body text)
- [ ] Spot-check the rebuilt appendix 05 heading links — TOC entries for Point 9, Point 11, Diseases 1/3/5 should still scroll to the right sections
- [ ] `npm test` still passes locally (55/55)

## Related issues

- Closes #164
- Closes #162
- Spawns #182 (brand-color contrast follow-up)
- Parent: #158

### DIFF
--- a/.github/workflows/accessibility.yml
+++ b/.github/workflows/accessibility.yml
@@ -10,6 +10,10 @@ on:
     branches:
       - main
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   pa11y:
     runs-on: ubuntu-latest
@@ -73,7 +77,7 @@ jobs:
     - name: Restore R packages
       shell: Rscript {0}
       run: |
-        if (!requireNamespace("renv", quietly = TRUE)) install.packages("renv", version = "1.0.3")
+        if (!requireNamespace("renv", quietly = TRUE)) install.packages("renv")
         renv::restore()
 
     - name: Setup Node.js
--- a/.github/workflows/accessibility.yml
+++ b/.github/workflows/accessibility.yml
@@ -1,0 +1,105 @@
+name: Accessibility
+
+permissions:
+  contents: read
+
+on:
+  pull_request:
+    types: [opened, synchronize]
+  push:
+    branches:
+      - main
+
+jobs:
+  pa11y:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
+
+    - name: Setup R
+      id: setup-r
+      uses: r-lib/actions/setup-r@6f6e5bc62fba3a704f74e7ad7ef7676c5c6a2590  # v2
+      with:
+        r-version: '4.5.1'
+        use-public-rspm: true
+
+    - name: Install system dependencies
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y \
+          libcurl4-openssl-dev \
+          libssl-dev \
+          libxml2-dev \
+          libpng-dev \
+          libjpeg-dev \
+          libcairo2-dev \
+          libfontconfig1-dev \
+          libfreetype6-dev \
+          build-essential \
+          gfortran \
+          libgmp3-dev \
+          libglpk-dev
+
+    - name: Setup Pandoc
+      uses: r-lib/actions/setup-pandoc@6f6e5bc62fba3a704f74e7ad7ef7676c5c6a2590  # v2
+      with:
+        pandoc-version: '3.1.11'
+
+    - name: Cache Quarto installation
+      id: cache-quarto
+      uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae  # v5.0.5
+      with:
+        path: /opt/quarto
+        key: quarto-1.4.549
+
+    - name: Install Quarto
+      if: steps.cache-quarto.outputs.cache-hit != 'true'
+      run: |
+        curl -LO https://github.com/quarto-dev/quarto-cli/releases/download/v1.4.549/quarto-1.4.549-linux-amd64.deb
+        sudo dpkg -i quarto-1.4.549-linux-amd64.deb
+
+    - name: Add Quarto to PATH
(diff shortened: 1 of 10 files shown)


lddurbin:
Thanks for the review! Addressed both high-value items in 0159b7f:

- **Concurrency group** — added. Rapid PR pushes will now cancel in-flight runs.
- **renv version pin** — dropped `version = "1.0.3"`. The fallback install now gets whatever `install.packages` resolves to, and `renv::restore()` bootstraps to the lockfile version (`1.2.0`) from there. Didn't touch `deploy.yml`'s identical line — happy to do a follow-up that harmonises both if you'd rather keep them aligned.

**Quarto version check** — both workflows pin `v1.4.549` identically, so rendered output is consistent. ✓

**`hideElements` vs per-selector rule ignore** — good point for #182 follow-up. pa11y-ci does support per-selector `ignore` entries via the object form, so when we fix those three contrast failures we can tighten the suppressions rather than drop-wholesale. Captured in #182.

**05-day-5.qmd heading style** — flagging for context: Pandoc only parses `{#id}` as a heading attribute when it's the **last** token on the line, so `*(page refs)*` must sit before the ID. The alternative would be to remove the trailing page-ref decoration, which changes the visible output. Keeping the current form — functionally correct, visual heading unchanged (the `{#id}` is stripped), and the duplicate-ID bug is gone.

PR #253, opened 2026-04-24
Add per-chapter reading-time indicator (#169)

### LINKED ISSUE #169: Show estimated reading time per chapter
## Problem

Readers benefit from knowing how long a chapter will take, especially those with cognitive fatigue who plan reading in blocks. The 12 Days already uses a "Day" pacing metaphor at the top level; per-chapter reading times are a finer-grained complement.

## Scope

- Compute reading time per chapter (word count ÷ 200 wpm is the standard).
- Display under each chapter heading: e.g. "≈ 8 min read".
- Implement as a Quarto Lua filter or a small R helper that runs at render time.
- Optional: include activity time separately ("≈ 8 min read + 10 min activity") to align with the existing timing-indicator convention.

- **Cost**: ~Half a day for the filter. Maintenance-free once written.
- **Benefit**: Cognitive accessibility — readers can plan sessions. Helps fatigued, time-constrained, and ADHD readers. Useful for everyone.
- **Visibility to default users**: **Small visible addition** — one line of small text under each chapter title.
- **Risk**: Could feel "blog-like" or conflict with the course's existing pacing language. Worth a sanity check before rolling out broadly — apply to a couple of chapters first and review.

## Acceptance

- Each chapter shows an estimated reading time under its heading.
- Calculation excludes code blocks, image captions, and admin sections.
- Style is consistent with the course's voice (not blog-feeling).

## Parent

#158

lddurbin:
## Summary

- Adds `filters/reading-time.lua`, a Pandoc Lua filter registered in `_quarto.yml`, that injects an estimated reading time under each chapter's H1.
- Counts prose words at 200 wpm, excluding `CodeBlock`, `RawBlock`, image captions, and Quarto's hidden nav/meta divs (which would otherwise inflate counts ~3×).
- Detects chapters that embed activity widgets (`viewof`, `Inputs.textarea`, `createDownloadButton`, or a `.thought` div) and appends "+ activities" to the label so the number isn't misread as total engagement time on activity-heavy pages.
- Chapters with fewer than 50 prose words omit the indicator rather than displaying a misleading "~ 1 min" on activity-only pages.
- Adds a muted-italic CSS rule for `.reading-time` (hidden in print).
- Documents the mechanism in `workflow/CONVERSION_GUIDE.md`.

## Why

Per issue #169, readers — especially those with cognitive fatigue or time constraints — benefit from knowing how long a chapter will take so they can plan reading in blocks. The existing per-chapter clock (`create_clock`) is a time-of-day pacing cue, not a duration, so the two indicators are complementary. The "+ activities" variant was added during review to avoid under-reporting engagement on activity-heavy chapters like Day 8.

## Test plan

- [x] `quarto render` completes with no warnings
- [x] All 120 chapters rendered; indicator present on every chapter above the 50-word threshold
- [x] Sample numeric spot-checks line up with `wc -w` against a 200 wpm baseline (e.g. 453-word coda → 3 min, 1825-word Deming (Re)discovered → 9 min)
- [x] Activity detection works: Day 8 `02-joy-in-work.qmd` (ojs-heavy) shows "+ activities"; Balaji coda (prose-only) does not
- [x] Layout verified on clock-column chapter (`day-01/01-overture.qmd`): indicator sits between H1 and `.columns` scaffold, does not collide with clock
- [ ] Reviewer to eyeball a few chapters in the preview to confirm the visual tone matches the course voice

Closes #169

### DIFF
--- a/_quarto.yml
+++ b/_quarto.yml
@@ -176,6 +176,9 @@ book:
     - content/appendix/contributions-balaji-reddie/07-indias-missed-opportunity.qmd
     - content/appendix/11-references-and-sources.qmd
 
+filters:
+  - filters/reading-time.lua
+
 format:
   html:
     theme: cosmo
--- a/assets/styles/main.css
+++ b/assets/styles/main.css
@@ -885,3 +885,15 @@ button:focus,
   .dyslexia-toggle { display: none !important; }
 }
 
+.reading-time {
+  margin-top: -0.5rem;
+  margin-bottom: 1.5rem;
+  font-size: 0.9rem;
+  color: #6c757d;
+  font-style: italic;
+}
+
+@media print {
+  .reading-time { display: none !important; }
+}
+
--- a/assets/styles/main.css
+++ b/assets/styles/main.css
@@ -889,7 +889,7 @@ button:focus,
   margin-top: -0.5rem;
   margin-bottom: 1.5rem;
   font-size: 0.9rem;
-  color: #6c757d;
+  color: var(--bs-secondary-color, #6c757d);
   font-style: italic;
 }
 
--- a/filters/reading-time.lua
+++ b/filters/reading-time.lua
@@ -68,10 +68,14 @@ local function count_blocks(blocks)
   for _, b in ipairs(blocks) do
     if b.t == "CodeBlock" or b.t == "RawBlock" or b.t == "HorizontalRule" then
       -- skip
-    elseif (b.t == "Div" or b.t == "Para" or b.t == "Plain") and has_class(b, "hidden") then
+    elseif b.t == "Div" and has_class(b, "hidden") then
       -- skip Quarto's injected navigation/meta blocks
     elseif b.t == "Para" or b.t == "Plain" then
       total = total + count_words(stringify_inlines(b.content))
+    elseif b.t == "LineBlock" then
+      for _, line in ipairs(b.content) do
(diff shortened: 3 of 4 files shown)


lddurbin:
Thanks for the thorough review — all five points addressed in 68e0398:

1. **Dead code**: narrowed the hidden-block skip to `Div` only.
2. **Table captions**: switched to a "stringify whole table, subtract caption words" approach. (First tried stringifying `b.head`/`b.bodies`/`b.foot` directly, but this Pandoc version doesn't accept TableHead/TableBody in `pandoc.utils.stringify` — it errors with "while retrieving stringifyable element". The subtraction approach handles both `.long` and `.short` captions and keeps the math bounded with `math.max(0, ...)`.)
3. **LineBlock**: added — iterates each line and counts via `stringify_inlines`.
4. **aria-label escape**: added a small `html_escape` helper and applied it to both `label` and `suffix`.
5. **Hardcoded colour**: switched to `var(--bs-secondary-color, #6c757d)` so theme swaps are respected, with the Bootstrap grey as a defensive fallback.

Sanity re-render: all 115 eligible chapters still produce an indicator, `day-02/06-your-turn.qmd` (the table-heavy chapter) now renders without error.

On the 200 wpm observation — noted as a deliberate choice; happy to make it configurable via a `WPM` metadata key later if we want to tune to a different baseline.

PR #256, opened 2026-04-24
Add light/dark mode toggle via dual Quarto theme (#165)

lddurbin:
## Summary

- Switch `_quarto.yml` from single `theme: cosmo` to `theme: { light: cosmo, dark: darkly }`, prompting Quarto to auto-inject a `.quarto-color-scheme-toggle` in the navbar. Light stays default; darkly is the alternate, so visitors whose OS prefers dark mode land on it automatically on first visit and the choice persists thereafter via `localStorage`.
- Audit every hard-coded colour in `assets/styles/main.css` and add a `body.quarto-dark`-scoped override section at the end of the file. Covers design tokens (`--color-bg-light`, `--color-callout-emphasis`, `--color-outcome`, `--color-funnel`, `--color-marble`, `--color-highlight-green`, `--color-border-gray`), the callout strips (return/arrival/info/callout-emphasis), Deming quote/blue text, the Foreman transcript block, thought & thought_commentary boxes, technical_aid, funnel-experiment table/status/grid/summary, the dyslexia toggle, the float-box, and the two inline `style="color: blue"` / `style="color:#c0392b"` patterns used across appendix pages (defeated with attribute selectors + `!important`, the minimum needed to beat inline styles).
- Extend the existing `@media (prefers-contrast: high)` rules so high-contrast users still get reinforced borders and quote colours in dark mode.
- Update `accessibility.qmd`: promote "Light/dark mode toggle" from the *What's still to come* list into *Current features*, and note the `prefers-color-scheme` honouring + cross-page persistence.

Closes #165.

## Why

The issue scope asked for a discoverable in-page toggle in addition to OS `prefers-color-scheme`, plus a readable experience across callouts, floating boxes, and the funnel-experiment widgets. Quarto's dual-theme feature handles the wiring and persistence for free; the bulk of the work was the custom-CSS audit.

## Test plan

- [ ] Verify the new toggle appears in the navbar on every page (home, a day chapter, funnel-experiment page, an appendix page).
- [ ] Toggle to dark mode and check legibility on:
  - [ ] a callout-heavy page (e.g. Day 1 or Day 2 chapters with `.return_callout` / `.arrival_callout` / `.info_callout`)
  - [ ] Day 2 Ch 7 (Foreman transcript block — entries, context labels, empty state)
  - [ ] a page with `.thought` / `.thought_commentary` boxes
  - [ ] Day 3 funnel-experiment page (data tables, status bar, comparison grid, summary table, buttons)
[...]

Issue #165, opened 2026-04-17
Add light/dark mode toggle via Quarto theme


lddurbin:
## Problem

Many readers prefer dark mode for low-light reading, photosensitivity, or simply preference. The browser exposes `prefers-color-scheme`, but a discoverable in-page toggle reaches users whose OS preference doesn't match their reading preference.

## Scope

- Switch `_quarto.yml` `theme: cosmo` to `theme: { light: cosmo, dark: darkly }` (Quarto's built-in dual-theme syntax). Quarto adds a toggle to the navbar automatically.
- Audit `assets/styles/main.css` custom colours for dark-mode contrast: `--color-bg-light`, `--color-callout-emphasis`, `.deming_blue`, `.deming_quote`, the funnel-experiment palette, and all hard-coded background colours in callouts.
- Use CSS variables and `[data-bs-theme="dark"]` selectors to override hard-coded colours where needed.
- Test with `prefers-contrast: high` so the existing high-contrast rules still work in dark mode.

## Cost / Benefit / Visibility

- **Cost**: Theme swap is trivial; the audit and dark-mode CSS overrides are the bulk. Probably ~half a day.
- **Benefit**: Dark-mode readers (very common preference). Photosensitivity. Pairs well with OpenDyslexic for some readers.
- **Visibility to default users**: **Small visible addition** — a toggle in the navbar. Default light mode is unchanged.

## Acceptance

- Quarto navbar toggle is present and functional on every page.
- All custom CSS colours have legible dark-mode equivalents (WCAG AA contrast).
- Funnel experiment, callouts, and floating boxes all readable in dark mode.
- Print stylesheet still produces a light-on-white print.

## Parent

#158

Issue #157, opened 2026-04-17
Implement dyslexia font toggle (CSS, JS, YAML wiring)


lddurbin:
## Problem

With the OpenDyslexic font assets in place (parent #155, prerequisite issue for assets), readers need a way to switch the site to OpenDyslexic and have the choice persist across pages. The toggle must be keyboard-accessible, FOUC-free on reload, and must not break code blocks or data tables.

### `assets/styles/main.css`

- Four `@font-face` rules (Regular / Bold / Italic / Bold-Italic) following the existing DM Serif Display pattern at `main.css:1-15`.
- A `body.dyslexic-font` rule that swaps the font on body, headings, paragraphs, lists, table cells, and callouts.
- An override that **preserves monospace** on `code`, `pre`, `.data-sequence-mono`, and `.fe-data-table` — OpenDyslexic kills code readability.
- A `.dyslexia-toggle` button style — fixed bottom-right, ≥44×44px (matches the mobile touch-target standard at `main.css:307-310`), hidden in print via `@media print`.

### `assets/scripts/font-toggle.js` (new file, plain script — not an ES module)

- On `DOMContentLoaded`: inject the toggle button into `<body>`, mirroring the skip-link injection at `_quarto.yml:163-169`.
- Button has `aria-label="Toggle dyslexia-friendly font"` and `aria-pressed` reflecting current state.
- Click handler flips `body.dyslexic-font`, updates `aria-pressed`, persists to `localStorage` under key `td:dyslexic-font`.

### `_quarto.yml` `header-includes`

- Add `<script src="/assets/scripts/font-toggle.js"></script>` alongside the existing `functions.js` line.
- Add an inline synchronous `<script>` that reads `localStorage` and adds `dyslexic-font` to `document.documentElement` before paint (FOUC prevention).
- CSS rule pattern: `html.dyslexic-font body, body.dyslexic-font { ... }` so the rule applies whether the class lands on `<html>` (early) or `<body>` (after toggle).

## Acceptance

- Toggle button appears bottom-right on every page, including the appendix.
- Click swaps the font immediately; `aria-pressed` updates accordingly.
- Reload retains the choice with **no flash** of the default font.
- Tab-focusable; Space and Enter activate it.
- Print preview hides the toggle and reverts to the print font.
- Code blocks (`.cell pre`, `code`) and the funnel-experiment data table (`.fe-data-table`) stay in monospace even when the toggle is on.
[...]

PR #258, opened 2026-04-24
Group reading preferences into a single panel (#166)

### LINKED ISSUE #166: Group reading preferences into a single panel
## Problem

If we add the OpenDyslexic toggle (#157), a light/dark toggle (#165), and possibly font-size or line-height controls later, the corner-button approach multiplies into clutter. Better to surface them under a single discoverable "Reading preferences" control.

## Scope

- A single icon button (gear or "Aa") opens a small panel containing all reading-preference toggles.
- Panel persists position; toggles persist values via `localStorage`.
- Consolidates the OpenDyslexic toggle (from #157) into this panel rather than letting it ship standalone.
- Easily extensible — adding line-height or font-size sliders later is a panel update, not a new corner button.

## Decision needed

Should this be done **before** #157 ships (so #157 builds the panel), or **after** (refactoring the standalone toggle into the panel)?

- **Before**: #157 scope grows; standalone toggle is never shipped.
- **After**: #157 ships fast, then this issue refactors. **Risk**: refactor never happens.

I lean toward "before" — refactors of shipped UI tend to be deferred indefinitely, and this changes #157's scope significantly.

- **Cost**: Slightly more JS than a single toggle. Roughly half a day on top of #157's existing scope.
- **Benefit**: Single discoverable a11y entry point. Future-proof for additional preferences.
- **Visibility to default users**: One small icon vs multiple toggles — net visual *reduction* compared to shipping each toggle as its own corner button.

## Acceptance

- One button opens a panel containing all reading-preference toggles.
- OpenDyslexic (#157), dark mode (#165), and any future toggles live here.
- Panel is keyboard-accessible (Esc closes, Tab cycles through controls).
- Panel state is announced to screen readers via `aria-expanded`.

## Parent

#158

## Related

#155, #157, #165, #170

lddurbin:
## Summary

Closes #166.

Consolidates the standalone OpenDyslexic toggle (#157) and Quarto's auto-injected navbar light/dark toggle (#165) into a single bottom-right **Reading preferences** trigger that opens a panel containing both controls. Future preferences (font size, line-height, etc.) drop into the panel as additional rows rather than new corner buttons.

### What changed

- **New script** `assets/scripts/reading-prefs.js` builds a fixed corner trigger and a hidden panel; replaces `font-toggle.js`.
- **Trigger** uses `aria-expanded` / `aria-controls`; **panel** uses `role="dialog"` with an `aria-label`.
- **Theme control** defers to `window.quartoToggleColorScheme()` so Quarto keeps managing stylesheet swaps, syntax-highlight CSS, and the `quarto-color-scheme` localStorage key — no logic duplication.
- **Dyslexia control** retains the existing `td:dyslexic-font` localStorage key + `dyslexic-font` body/html class so persistence is unchanged.
- **Keyboard**: Esc closes, click-outside closes, focus moves into the panel on open and returns to the trigger on close.
- **Quarto's navbar toggle** is hidden via `.quarto-color-scheme-toggle { display: none !important }` so the panel is the single entry point. The link still emits in HTML so its handler stays wired for our Theme control to invoke.
- **Dark-mode + HC overrides** cover all panel surfaces (toggle, panel bg, title, row separator, controls in both pressed/unpressed states).
- **`accessibility.qmd`** updated: rewrites the two separate "OpenDyslexic font toggle" and "Light/dark mode toggle" bullets into a single "Reading preferences panel" entry with sub-bullets for the current toggles.

### Why this approach

Driving Quarto's existing API rather than reimplementing the stylesheet swap means the syntax-highlighting CSS, Bootstrap variant, and color-scheme event all keep working without us reinventing them. Hiding (rather than removing) Quarto's navbar `<a>` keeps `window.quartoToggleColorScheme` available as a global for our panel to call.

## Test plan

- [x] `quarto render` succeeds across all 120 chapters
- [x] `pa11y-ci` passes — 29/29 URLs, 0 errors (the prior duplicate-id ignore from #165 keeps the dual-stylesheet `id` collision suppressed; no new ignores added)
- [x] Headless browser smoke test (puppeteer) verifies:
  - Trigger and panel render; navbar toggle is hidden via CSS
  - Click trigger → `aria-expanded="true"`, panel visible, focus inside panel
  - Click Theme → `body.quarto-dark` toggles, button state updates ("Light" ↔ "Dark"), `localStorage["quarto-color-scheme"]` updates
  - Click Dyslexia font → `body.dyslexic-font` toggles, button state updates ("Off" ↔ "On"), `localStorage["td:dyslexic-font"]` updates
  - Esc closes panel and returns focus to trigger
  - Click outside closes panel
- [ ] Manual: confirm panel collapses to icon-only at <700px viewport
- [ ] Manual: confirm panel readable and high-contrast in dark mode + `prefers-contrast: high`
- [ ] Manual: print preview confirms trigger and panel are suppressed

## Related

- Parent: #158
- Consumed: #157 (OpenDyslexic toggle), #165 (light/dark toggle, merged in PR #256)


lddurbin:
Thanks for the review — addressed in a follow-up commit.

**Issue 1 — HC overrides (agreed, fixed).** Added `reading-prefs-*` rules to both `prefers-contrast: high` blocks (`main.css:601` and `main.css:1262`). Used thicker borders in light-HC and `currentColor`-based outlines on pressed controls so Windows HC palettes can override cleanly; added matching dark-HC border-color overrides.

**Issue 2 — focus trap (different fix).** Dropped `role="dialog"` rather than adding a focus trap. The panel is a non-modal disclosure — outside clicks dismiss, no backdrop, nothing below is inert — so the ARIA APG disclosure pattern fits better than the dialog pattern. The trigger already carries `aria-expanded`/`aria-controls`; I switched the panel to `aria-labelledby` pointing at the title. That removes the dialog contract (focus trap, `aria-modal`) we weren't honouring anyway. Happy to reconsider if you'd prefer the trap version.

**Issue 3 — theme state race (pushing back).** `quartoToggleColorScheme` is fully synchronous in Quarto 1.4–1.8: it calls `toggleColorMode` → `toggleBodyColorMode`, which mutates `body.classList` inline (see `_book/about-this-edition.html:261-268` + `:115-123`). So reading `darkIsActive()` immediately after is correct.

The suggested fallback — `if (on === wasDark) on = !wasDark` — would actively misbehave when `window.quartoToggleColorScheme` is unavailable (we guard on `typeof === "function"` at line 30, so `toggleDark()` is a no-op). In that case the button would flip its `aria-pressed` state without the stylesheet changing, leaving the label desynced from the actual theme. That's a worse failure mode than the hypothetical deferred-mutation case it's defending against. Leaving as-is; if a future Quarto version defers the mutation we can add a `MutationObserver` on `body.class` then.

**Minor — h2 in outline (agreed, fixed).** Replaced `<h2 class="reading-prefs-title">` with `<p class="reading-prefs-title" id="reading-prefs-title">` and wired it as the panel's `aria-labelledby` target. Existing `.reading-prefs-title` styles carry over.

**Minor — snap vs animated yield (skipping).** Agree with your own call that it's fine in practice; the footer is reached by scroll only, and the panel being open + user already at footer is a narrow window.

Issue #261, opened 2026-04-25
Glossary content audit — pick ~10 course-defining terms to mark up


lddurbin:
## Problem

#167 landed the `<dfn data-definition="…">` mechanism with three terms marked up on Day 1 as proof. The natural follow-up is "now mark up everything that recurs" — but doing that uncritically is **worse than doing nothing**. Three costs apply once the underline is sprinkled across many chapters:

1. **Tab-stop pollution.** Every `<dfn>` is a keyboard tab stop. Forty marked terms across the book = forty extra hops for keyboard users — exactly the population we're trying to help.
2. **Visual noise.** Dotted underlines on every key term turn pages into a Christmas tree and constantly interrupt linear reading flow.
3. **Pedagogical short-circuiting.** Neave teaches the vocabulary deliberately. A tooltip can resolve "what does this mean?" before the surrounding prose has done its job.

The win for tooltips is concentrated in **non-linear and returning readers** — someone reading Day 7 who's forgotten what "special cause" means a week after Day 3. That's a real but small audience, and the marginal benefit drops quickly past the most-recurring terms.

## Scope

Mark up **at most ~10** course-defining terms. The discipline is the cap, not the floor. Each term added must justify its place against the costs above.

Candidate list (to be ratified or trimmed during the audit):

- common cause / common-cause variation
- special cause / special-cause variation (note: Shewhart's "assignable" cause is a synonym — should it share the dfn?)
- in / out of statistical control
- PDSA cycle (Shewhart cycle)
- the Joiner triangle
- the funnel experiment
- system of profound knowledge
- appreciation for a system
- operational definition
- transformation (in the Deming-specific sense, not the everyday sense)

Where the term first appears in the prose, mark up that one defining instance — **per the rule already documented in `workflow/CONVERSION_GUIDE.md` § Glossary tooltips for specialised terms.** Subsequent uses stay plain.

## Definition sourcing

Definitions come from **Neave's own prose**, not from editorial paraphrase. For each term:

1. Locate the sentence(s) where Neave most cleanly defines or characterises the term, somewhere in the book.
2. Tightly paraphrase or quote into a one-or-two-sentence `data-definition` value.
3. Cite the source location in the audit checklist (so future-us can verify).

If Neave doesn't actually give a clean definitional sentence anywhere — only describes the term across many paragraphs — that term is a **bad candidate** for a tooltip and should be dropped from the cap rather than synthesised.

## Cost / Benefit / Visibility

[...]

PR #262, opened 2026-04-25
Build dfn glossary tooltip mechanism (#167)

### LINKED ISSUE #167: Build dfn tooltip mechanism for glossary terms
## Problem

The 12 Days uses heavy specialised vocabulary — "common cause", "special cause", "PDSA cycle", "system of profound knowledge", "appreciation for a system", "operational definition". Readers new to the field have to look these up externally. A consistent in-page glossary mechanism reduces cognitive load.

This issue covers the **mechanism only**; the content audit (identifying and marking up all terms) is a separate issue.

## Scope

- Define a markup pattern: `<dfn data-definition="…">term</dfn>` (or similar Quarto-friendly equivalent).
- CSS: subtle visual cue (dotted underline) on `dfn[data-definition]`.
- JS: on hover/focus, show a tooltip with the definition. Must be keyboard-accessible — focus shows tooltip, Esc dismisses.
- Use `aria-describedby` (not `title`) so screen readers announce the definition correctly.
- Document the pattern in `workflow/CONVERSION_GUIDE.md`.
- Mark up one example chapter (e.g. Day 1) as proof.

- **Cost**: ~Half a day for JS + CSS. Tooltip a11y is the tricky part — must work with screen readers, not just mouse hover.
- **Benefit**: Cognitive accessibility for newcomers, ESL learners, returning readers who forget terms.
- **Visibility to default users**: **Subtle** — a dotted underline on glossary terms tells the reader "this is a defined term you can hover for context". A small positive change.

## Acceptance

- `<dfn>` markup with definition tooltip works on hover **and** on keyboard focus.
- Screen readers announce both the term and the definition.
- Pattern documented in `CONVERSION_GUIDE.md`.
- One example chapter uses the pattern.

## Parent

#158

lddurbin:
## Summary

- Adds the `<dfn data-definition="…">term</dfn>` markup pattern. The accompanying `assets/scripts/dfn-tooltip.js` makes each dfn keyboard-focusable, injects a child `<span role="tooltip" aria-hidden="true">` carrying the definition, and wires `aria-describedby` so screen readers announce the definition automatically. Reveal is pure CSS via `:hover` / `:focus`; pressing Esc dismisses for keyboard users.
- Documents the pattern in `workflow/CONVERSION_GUIDE.md` (when to mark up, when not to, definition style) and announces the feature in `accessibility.qmd`.
- Marks up three terms on Day 1 (`11-deming-story.qmd`) as proof: *common causes*, *special causes*, *PDSA cycle*. Definitions are sourced from Neave's own prose (the same chapter for the cause types; Day 11 ch. 3 for PDSA), not editorial paraphrase.
- Adds `11-deming-story.html` to the pa11y URL list so the new pattern gets WCAG-checked in CI.

The follow-up content audit — picking the ~10 course-defining terms to mark up — is filed as #261, with the cap baked into its acceptance criteria.

Closes #167.

## Test plan

- [x] Quarto render succeeds; rendered HTML contains all three dfns and the script include.
- [x] JS passes `node --check`.
- [x] Visual review on Day 1 → "The Deming Story" page, both light and dark mode:
  - dotted underline visible on the three terms
  - help cursor on hover
  - tooltip appears just above the term on hover and on Tab focus (parity)
  - Esc dismisses the focused tooltip
  - dark-mode tooltip stays distinguishable on darkly's `#222` body
- [ ] pa11y CI passes against the updated URL list (will run on PR).

### DIFF
--- a/.pa11yci.json
+++ b/.pa11yci.json
@@ -36,6 +36,7 @@
     "http://localhost:8765/content/days/day-01/01-overture.html",
     "http://localhost:8765/content/days/day-01/02-rediscovered.html",
     "http://localhost:8765/content/days/day-01/07-quality-theory.html",
+    "http://localhost:8765/content/days/day-01/11-deming-story.html",
     "http://localhost:8765/content/days/day-01/13-major-activity.html",
     "http://localhost:8765/content/days/day-02/01-run-charts.html",
     "http://localhost:8765/content/days/day-02/04-our-first-control-chart.html",
--- a/_quarto.yml
+++ b/_quarto.yml
@@ -198,6 +198,7 @@ format:
       </script>
       <script src="/assets/scripts/functions.js"></script>
       <script src="/assets/scripts/reading-prefs.js"></script>
+      <script src="/assets/scripts/dfn-tooltip.js"></script>
       <script>
         document.addEventListener("DOMContentLoaded", function () {
           var a = document.createElement("a");
--- a/accessibility.qmd
+++ b/accessibility.qmd
@@ -19,6 +19,8 @@ The panel currently contains:
 - **Theme** — switches the site between light and dark colour schemes for low-light reading, photosensitivity, or simple preference. Honours your operating system's `prefers-color-scheme` setting on first visit and remembers your explicit choice thereafter.
 - **Dyslexia font** — switches body and heading text to the [OpenDyslexic](https://opendyslexic.org/) typeface, which some readers with dyslexia find easier to parse.
 
+**Glossary tooltips on specialised terms.** Specialised vocabulary (such as *common cause*, *special cause*, and *PDSA cycle*) is marked up with the HTML `<dfn>` element and given a subtle dotted underline. Hovering or keyboard-focusing the term reveals a short definition. Screen readers announce the definition automatically via `aria-describedby`, and pressing `Esc` dismisses the tooltip.
+
 **Visible focus indicators.** Keyboard focus is clearly outlined on all interactive controls (links, buttons, text inputs, toggles). Mouse users are unaffected.
 
 **ARIA labelling.** Interactive elements, data tables, and input widgets expose accessible names to assistive technology. This includes the site search, the activity text inputs, and the chart-embedded controls on Days 2, 3, and 6.
--- a/assets/scripts/dfn-tooltip.js
+++ b/assets/scripts/dfn-tooltip.js
@@ -24,14 +24,18 @@
     tip.id = tipId;
     tip.setAttribute("role", "tooltip");
     // The tooltip is a child of the dfn so the dfn's `position: relative`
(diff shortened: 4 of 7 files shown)


lddurbin:
Thanks for the careful review. Pushed `25a948b` addressing issues 1–3:

**Issue 1 (`aria-hidden`):** agreed and removed. The `visibility: hidden` default already keeps the tooltip out of the a11y tree until reveal, and dropping `aria-hidden` side-steps the NVDA/older-VoiceOver edge case you flagged. Updated the code comment to explain why visibility alone is sufficient.

**Issue 2 (right-edge overflow):** took your `left: 50%; transform: translateX(-50%)` fix. Centering improves the average case noticeably; terms very near either margin will still clip on narrow viewports, but a JS-based `getBoundingClientRect` clamp would be overkill while we're at three marked-up terms. Noted the residual gap inline so the next person reading the CSS knows the next escalation.

**Issue 3 (touch users):** real gap, but I disagreed with the suggested fixes for this PR's scope:

- A separate tap-toggle state machine doubles the script's complexity and overlaps confusingly with the hover/focus reveal path that's already wired.
- Linking to a glossary page presupposes a glossary page that doesn't exist yet — that's part of the #261 content audit.

Took a smaller third option: explicit `dfn.focus()` on click. iOS Safari has historically been inconsistent about tap-focusing non-button `tabindex=0` inline elements; forcing focus on click makes the existing `:focus` reveal path work on tap without a new state machine. Genuine touch-first UX (long-press, dismiss-on-tap-outside, etc.) is a follow-up — touch deserves its own design pass and is out of scope for the mechanism PR.

**Minor notes:**

- `var` vs `const`/`let` — the IIFE pattern with `var` is the project's existing convention in `functions.js`-style scripts targeting older browsers; sticking with it keeps the asset script style consistent. Happy to bulk-modernise across all four asset scripts in a separate housekeeping PR if you'd prefer.
- `pointer-events: none` blocking text selection — accepted as a v1 trade-off; revisit if definitions grow longer.

lddurbin:
Pushed `4ae166e`:

**Issue 1 (asymmetric fade):** taken your fix — `transition: opacity 120ms ease, visibility 0s linear 120ms` on the default state, with `... linear 0s` on the reveal state so the visibility flip happens immediately on entry but is delayed on exit. Code comment explains why. Reduced-motion users still get the global `transition-duration: 0.01ms !important` override, so this doesn't regress that path.

**Issue 2 (top-of-viewport clipping):** deferred per your note that it's an acceptable defer. Added a one-line "Positioning caveat" in `workflow/CONVERSION_GUIDE.md` so editors marking up future terms know to avoid first-paragraph terms. The proper fix (JS-positioning or the Popover API with auto-flip) lands alongside any term that genuinely needs to live in a chapter opening — none of the candidates in #261 do.

Issue #239, opened 2026-04-22
Replace or remove dead external links (deming.org.uk, rqoq.org.nz)


lddurbin:
## Context

Surfaced by the #139 link-check sweep. Two external hosts return connection errors and need an editorial decision (remove, link to archive.org snapshot, or substitute).

- **`deming.org.uk`** (Deming Transformation Forum UK) — 3 references:
  - \`welcome.qmd\` (linked from the 'A. PLEASE START HERE' flow — high visibility)
  - \`content/days/day-01/02-rediscovered.qmd:158\`
  - \`content/days/day-01/06-outline.qmd:193\`
  - \`content/appendix/11-references-and-sources.qmd:140\`
- **`rqoq.org.nz`** (NZ Organisation for Quality) — 1 reference:
  - \`welcome.qmd:129\` (part of 'the PDF files may be viewed at…' list)

## Verification

\`\`\`
$ curl -sL -o /dev/null -w '%{http_code}\n' --max-time 20 https://www.deming.org.uk/
000
$ curl -sL -o /dev/null -w '%{http_code}\n' --max-time 20 https://www.rqoq.org.nz/
000
\`\`\`

Other Deming-related hosts listed in the same paragraphs (\`demingalliance.org\`, \`priscillapetty.com\`) are live — those were linkinator false positives, not in scope here.

## Decisions required

1. **`deming.org.uk`** — does the Forum still exist under another domain? If not: remove references or substitute with the Deming Alliance / NZOQ equivalent? Neave himself may know.
2. **`rqoq.org.nz`** — NZOQ's current URL is likely \`nzoq.org.nz\` (not verified here). Replace or drop?

## Acceptance

- [ ] Each of the 4 source references either resolves to a live URL, is removed, or is preserved as plain text with a note
- [ ] \`quarto render\` warnings remain clean
- [ ] External-link spot-check after the change

## Source

Found during #139 final link-check sweep.

Issue #213, opened 2026-04-21
Decide whether to strip inline [WB NNN] cross-references from day and appendix prose


lddurbin:
Spawned from #210 scope audit (paired with #212).

## Context

Neave's source material uses a dual page-referencing system throughout the day and appendix prose: when citing a page, he gives both the Day-chapter page number and the corresponding printable-Workbook page number, e.g.

> "See the 10th of the 14 Points on Day 5 page 8 *[WB 76]*."

A grep across \`content/\` turns up **54 occurrences across 26 files**. Each one is a mechanical drop of the \`*[WB NNN]*\` / \`[WB NNN]\` suffix, leaving the Day-chapter page reference untouched.

## The call to make

These refs differ from the appendix \`(Return to Workbook page X / Day Y page Z.)\` parentheticals that were stripped in the #210 PR. The appendix return-refs were full navigation parentheticals that would confuse a site-only reader. The inline \`[WB NNN]\` bracket suffixes don't mislead — they're clearly bracket-suffixed — and they retain cross-reference utility for readers who also have the NZOQ PDF edition open alongside the site.

So the question is: do we strip them for consistency with #185's Workbook removal stance, or preserve them as a cross-reference aid for PDF-edition readers?

## Scope if we strip

- [ ] Regex-style drop of \`*[WB NNN]*\` and \`[WB NNN]\` across 26 files in \`content/\`.
- [ ] Handle paired constructs like \`page 14--29 *[or WB 220--234 along with today's page 23]*\` where the whole bracketed alternative is Workbook-specific.
- [ ] Deviations-log entry in \`docs/deviations-from-source.md\`.

## Depends on

- #211 ✅ deviations-log format locked.
- Coordinate with #212 if both land together — one log entry can cover both.

Issue #212, opened 2026-04-21
Rewrite index.qmd front-matter Workbook prose


lddurbin:
Spawned from #210 scope audit (paired with #208).

## Context

`index.qmd` is the book's front matter (Preface / Active learning / Time management / Some important practical matters / Twelve Days? / etc.). Three of its sections carry ~34 references to the printed Workbook:

- **"To print or not to print…"** (around line 327) — frames the Workbook as the recommended printable deliverable
- **"Information boxes, etc"** (around line 409) — explains the `.workbook_callout` yellow info-box convention (the divs themselves were removed in #196/#202, but the explanatory prose survives)
- **"Guidance on printing and binding"** (around line 582) — instructions for physically assembling the four-part Workbook

This is structurally the twin of #208's appendix Welcome Workbook passage: Neave's original explanation of how to use the printed Workbook, which no longer applies to this site's in-page interaction model.

## Scope

- [ ] Rewrite or remove the Workbook-specific passages in each of the three sections above.
- [ ] Preserve the surrounding pedagogical points (print-vs-screen guidance, yellow/green info-box conventions, practical matters) where they generalise beyond the Workbook.
- [ ] Add a deviations-log entry in \`docs/deviations-from-source.md\`; this is material and significant — similar treatment to the #208 entry.

## Out of scope

- `content/appendix/00-welcome.qmd` — covered by #208.
- Appendix return-navigation refs (\`(Return to Workbook page X / Day Y page Z.)\`) — mechanical, handled in the #210 PR.

## Depends on

- #211 ✅ deviations-log format locked.
- Coordinate with #208 if both land together — single log entry can cover both if timed right.

PR #263, opened 2026-04-25
Apply glossary tooltip audit (#261)

### LINKED ISSUE #261: Glossary content audit — pick ~10 course-defining terms to mark up

lddurbin:
## Summary

Applies the editorial discipline that #261 was filed to enforce: a capped, sourced set of `<dfn data-definition="…">` tooltips, with the audit trail captured in a canonical document.

- **4 terms added**: "in statistical control" (Day 1 ch 11), System of Profound Knowledge (Day 1 ch 11), operational definition (Day 11 ch 4), transformation (Day 12 ch 4). Brings total to **7 marked terms** (3 from #167 + 4 here).
- **3 candidates rejected with rationale**: Joiner triangle (definition is carried by an adjacent diagram, not a sentence), funnel experiment (Neave describes the apparatus but the conceptual takeaway is taught experientially), "appreciation for a system" (Neave never tightly defines the phrase as a unit — only Deming's own definition of "system" appears in the chapter).
- **Cap discipline preserved**: 7/10 slots used, deliberately leaving headroom rather than padding. The cap is in `docs/glossary-audit.md` as the explicit thing to push back against, per #261's acceptance criteria.
- **Sourcing rule honoured**: every `data-definition` value comes from Neave's prose (or a Deming quote that Neave reproduces verbatim) — no synthesis from general knowledge.
- **`workflow/CONVERSION_GUIDE.md`** updated to point editors at the audit log before adding new dfns; the example list swaps "appreciation for a system" (now rejected) for "transformation".
- **`.pa11yci.json`** picks up Day 11 ch 4 and Day 12 ch 4 so the new dfns get WCAG-checked in CI. Day 1 ch 11 was already on the list from #167.
- **`docs/deviations-from-source.md`** entry summarises the audit per project log convention.

Closes #261. Companion to #262 (which introduced the mechanism in #167).

## Test plan

- [ ] `quarto render` succeeds for the four edited chapters (verified locally — Day 1 ch 11, Day 11 ch 4, Day 12 ch 4 all rendered cleanly; rendered HTML contains 5/1/1 `<dfn>` elements respectively, matching expectations).
- [ ] pa11y CI passes against the four chapters (Day 1 ch 11 was already on the URL list; Day 11 ch 4 and Day 12 ch 4 added in this PR).
- [ ] Spot-check in browser: hover/focus on each new dfn reveals the tooltip; Esc dismisses; tooltip wording reads as Neave's voice, not editorial paraphrase.
- [ ] Confirm no dfn appears inside `<span class="deming_quote">` (the convention from #167) — manually verified during edit.
- [ ] Confirm no dfn sits in the first paragraph of any chapter where the upward-opening tooltip would clip (positioning caveat in CONVERSION_GUIDE.md) — Day 11 ch 4 line 21 sits well below the heading; Day 12 ch 4 line 14 ditto; Day 1 ch 11 additions are deep in the chapter.


lddurbin:
Thanks — three quick verifications.

**1. Italic inside `<dfn>` (Day 11 ch 4, line 21).** Confirmed: Pandoc processes the Markdown italic correctly. Rendered HTML at `_book/.../04-theory-of-knowledge-operational-definitions.html`:

```html
<dfn data-definition="…"><em>operational</em> definition</dfn>
```

So the visible word renders as *operational* (italic), not literal asterisks. Behaviour is consistent with Pandoc's `markdown_in_html_blocks`/`markdown` extension defaulting to *on* for the Quarto HTML format.

**2. `<dfn>` inside a blockquote (Day 12 ch 4, line 14).** Verified the CSS positioning chain — no clipping risk in this chapter:

- The tooltip uses `position: absolute`; its containing block is the `<dfn>` itself (which has `position: relative`). The surrounding `<blockquote>` is *not* a positioned ancestor and has no `overflow: hidden` or `clip-path` applied (`grep -nE "overflow|clip" assets/styles/main.css` shows only `.visually-hidden` and table-scroll containers carry overflow rules — none in the blockquote chain).
- The tooltip opens upward and spills over the preceding paragraph, which is the normal behaviour for absolutely-positioned tooltips and is what we want.
- The clip risk identified by `CONVERSION_GUIDE.md`'s positioning caveat is **viewport top**, not blockquote top — and the Day 12 ch 4 dfn sits well below the chapter heading (line 14 in source, several paragraphs of intro precede the blockquote in the rendered page).

I think the existing first-paragraph caveat covers this case correctly without needing a separate blockquote rule. Happy to extend it if a future audit candidate hits the actual viewport-top edge inside a blockquote.

**3. `deviations-from-source.md` "Landed in" field.** Yes — the "Pending — tracked in #261" form is the project convention for entries committed before merge (matches the existing `#212`, `#213`, `#239` entries on the file). It gets bumped to the PR number on merge as part of the standard flow.

Issue #271, opened 2026-04-26
Manifest debt: download-button flags out of sync on Days 2, 10, 11


lddurbin:
The `scripts/check-structure.sh` validator emits 9 warnings for chapters where the QMD file genuinely contains a `createDownloadButton(...)` call but the corresponding `workflow/validation/day-NN-manifest.yml` entry still has `has_download_button: false`. They are warnings rather than failures, so they have accumulated silently — none was introduced by this work, but the audit done while shipping #160 surfaced them in one place.

## Affected chapters

**Day 2** (`workflow/validation/day-02-manifest.yml`):
- `06-your-turn.qmd`
- `07-some-recollections.qmd`
- `08-major-activity.qmd`

**Day 10** (`workflow/validation/day-10-manifest.yml`):
- `02-appreciation-for-a-system.qmd`
- `04-theory-of-variation.qmd`

**Day 11** (`workflow/validation/day-11-manifest.yml`):
- `02-theory-of-knowledge-prediction.qmd`
- `03-theory-of-knowledge-theory-and-learning.qmd`
- `04-theory-of-knowledge-operational-definitions.qmd`
- `06-knowledge-of-psychology.qmd`

## How to verify

```bash
for d in 2 10 11; do ./scripts/check-structure.sh "$d" 2>&1 | grep -B2 "WARN download"; done
grep -l "createDownloadButton" content/days/day-02/06-your-turn.qmd ...
```

Each flagged file does in fact contain a `createDownloadButton(...)` call, so the manifest is the side that is wrong — the fix is to flip `has_download_button: false` → `true` in the relevant entries.

## Acceptance criteria

- `./scripts/check-structure.sh N` for N in 2..12 emits zero warnings about download buttons.
- No QMD content changes; only manifest entries updated.

## Notes

- Most likely cause is that download buttons were added to chapters in later PRs without bumping the manifest at the same time (the same shape of debt that the Day 12 heading drift had until #160 fixed it).
- Worth also auditing the inverse direction (manifest says `true` but chapter has no button) — a quick grep across all manifests vs `createDownloadButton` would catch any.

PR #270, opened 2026-04-26
Add long-descriptions to charts in Days 2-3 (#159)

lddurbin:
## Summary

- Pairs every pedagogically-loaded chart, histogram, control chart, and diagram in Days 2 and 3 with a collapsed `<details><summary>Describe this chart</summary>…</details>` block. Screen-reader users and chart-novices now get the *what to see* lesson that alt text alone can't carry; default readers see only the small disclosure under each image.
- Documents the pattern (with conventions, syntax, and skip-rules) in `workflow/PATTERNS.md` next to the existing Image with Lightbox section, so #160 can roll the same convention out to the remaining days.
- Repetitive series (e.g. Day 2's progressive run charts, the 24 `your-turn` practice tables) get one description that names the pattern rather than 24 near-duplicates. Decorative images (apparatus illustration) and pure data tables (the weekly statisticians' tables in `07-some-recollections.qmd`) are skipped per the issue's acceptance criteria.

## Adjacent fix

Bumps `viewof_count` for `07-some-recollections.qmd` in `workflow/validation/day-02-manifest.yml` from 12 → 16 — manifest had been out of sync since #254 added the Spaniards' textareas. One-line clean-up; the structure check now passes for Day 2.

## Closes

Closes #159.

## Test plan

- [x] `quarto render` succeeds for every modified file in Days 2 and 3 (14 files)
- [x] Spot-checked rendered HTML — `<details><summary>` markup and inner markdown render as expected
- [x] `bash scripts/check-structure.sh 2` — all 59 checks pass (was 1 fail)
- [x] `bash scripts/check-structure.sh 3` — all 89 checks pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #272, opened 2026-04-26
Add long-descriptions to charts in Days 1, 4-12 (#160)

### LINKED ISSUE #160: Add long-descriptions to charts in remaining days
## Problem

Once the long-description pattern is established for Days 2-3 (#159), the remaining days need the same treatment. Days 1, 4-12, and the appendix have varying chart density — Days 4, 6, 7, 10, and 11 are the chart-heaviest of the rest.

## Scope

- Apply the long-description pattern (defined in #159) to all chart and diagram images in Days 1, 4-12, and the appendix.
- Keep alt text concise (≤120 chars); put the pedagogical description in the `<details>` block.

- **Cost**: Editorial — possibly 50+ images. Best done day-by-day during normal review rather than as one giant pass.
- **Benefit**: Same pedagogical accessibility as #159, applied to the rest of the course.
- **Visibility to default users**: **None** — `<details>` is collapsed by default.

## Acceptance

- All chart/diagram images across Days 1, 4-12, and the appendix have long descriptions.
- Pattern remains consistent with the one established in #159.

## Parent

#158

## Depends on

#159 (pattern must be established first)

lddurbin:
## Summary

Closes #160. Applies the long-description pattern established for Days 2-3 in #159 to every remaining day in the course.

- **21 `<details>` blocks added** across 14 chapters in Days 1, 4, 6, 7, 8, 9, 10, 11, 12
- **26 images correctly skipped** per the documented rules in `workflow/PATTERNS.md` (decorative portraits, title cards, blank apparatus, repetitive series, content available verbatim elsewhere)
- **Day 5 had no qualifying images** (only a misleading-alt decorative separator)
- **Bonus fix**: Day 12 manifest debt — the four activity headings `ACTIVITY 12-a` Parts 1/2/3 and `ACTIVITY 12-b` were present in the chapters but missing from the manifest, so the structure check was failing pre-existingly. Same approach as the Day 2 manifest fix in #270.

The repetitive-series rule did meaningful work in Day 8: Tables 1-5 appear twice (once full-page, once as inline insets), so 10 candidate images collapse to 2 descriptions — one anchoring each variant.

## Test plan

- [x] `quarto render` on each of the 14 touched files — all rendered cleanly with no parse warnings
- [x] `./scripts/check-structure.sh N` for N in 1..12 — all pass; only pre-existing download-button warnings remain (raised as #271)
- [x] Coverage sweep: 47 candidate images audited, every unpaired image matches an explicit skip rule
- [ ] Reviewer to spot-check disclosure rendering on a sample chapter (Day 1 `11-deming-story`, Day 9 `02-a-system`, Day 12 `03-wisdom-from-peter-scholtes` are good ones — they cover grViz, static diagram, and the satirical "System of Profound Trouble" parody)

## Follow-ups

- #271 tracks the remaining manifest debt (download-button flags out of sync on Days 2, 10, 11) which surfaced during this audit but is out of scope here.
- Once #271 is merged, `H37` (img-alt) suppression in `.pa11yci.json` and `docs/pa11y-ignores.md` line 33 can be re-audited — the long-descriptions cover the pedagogical alt-text gap that suppression was working around.

### DIFF
--- a/content/days/day-01/11-deming-story.qmd
+++ b/content/days/day-01/11-deming-story.qmd
@@ -322,6 +322,13 @@ grViz("
 ")
 ```
 
+<details>
+<summary>Describe this diagram</summary>
+
+A left-to-right flow diagram of a whole production system. Four supplier nodes (A, B, C, D) feed materials into **Receipt and test of materials**, which passes them to **Production, assembly, inspection**, then to **Distribution**, and finally to **Consumers**. Crucially, the diagram is *not* a one-way assembly line: **Consumer research** loops back into **Design and redesign**, which in turn feeds Receipt, Production, and Distribution; **Tests of processes, machines, methods, costs** also feeds back into Receipt, Production, and Distribution. The lesson is that quality is built by treating the whole organisation — suppliers, production, distribution, consumers, design, and ongoing tests — as a single interconnected system, with information flowing back from the consumer end as much as forward.
+
+</details>
+
 <span class=deming_quote>
 2. Quality is determined by the management. Outgoing quality can not be better than the intentions of the management.</span> (Do you remember “Quality is made at the top” from the Overture?)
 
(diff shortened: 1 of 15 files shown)


lddurbin:
Thanks — both points addressed:

**Placement fix (1)**: You're right, the rendered HTML confirmed the disclosure ended up outside the `column-margin column-container` wrapper, leaving it in the main column while the image floated into the margin gutter. Pushed [a6c2b14](https://github.com/lddurbin/twelve_days_to_deming/commit/a6c2b14) which moves `<details>` inside the `:::` block. Both image and disclosure now render together as a single margin unit. Verified in the regenerated `_book/content/days/day-08/06-major-activity.html`.

**Repetitive-series question (2)**: Holding the line on no stub disclosures for Tables Two–Five. Reasoning:

- The repetitive-series rule documented in `workflow/PATTERNS.md:336-338` says "describe the first and note that subsequent charts add one point each. Don't restate."
- That's exactly what was applied in the precedent on PR #270 — Day 2's progressive run charts (`01-run-charts.qmd`) and the 24-red-beads charts (`06-your-turn.qmd`) each got one description on the first chart and zero on the rest, with no back-references.
- The Table One description already carries the explicit pointer-back: "Tables Two through Five reappear as similar margin insets later in the chapter, each showing the corresponding worked stage of the exercise; the layout convention is identical to Table One." A screen reader hitting the Table Two image will get the same alt text it would for any of the five (`Table Two — DemDim example`); the convention is that pedagogically-loaded *layout* gets described once.

If we want to revisit the repetitive-series rule itself — e.g. require a 1-line "see Table One disclosure for layout" stub on every follower — that's a worthwhile change but should happen at the PATTERNS.md level and apply retroactively to #270's series too. Happy to file a separate issue if you'd like that pursued; otherwise this PR follows the established convention.

PR #276, opened 2026-04-27
Wire check-structure.sh into CI; promote drift WARNs to FAIL (#274)

### LINKED ISSUE #274: Wire check-structure.sh into CI and promote download-button drift to FAIL
## Context

#271 / #273 fixed 10 download-button manifest entries that had drifted out of sync with the QMD source files over multiple PRs. The drift was silent because:

1. **`scripts/check-structure.sh` is not wired into CI.** The script header at `scripts/check-structure.sh:16-18` notes \"CI (Ubuntu) does not install Ruby by default\" — the validator only ever runs locally when someone remembers to invoke it.
2. **Drift is reported as `WARN`, not `FAIL`.** Per `scripts/check-structure.sh:351`, a download button found in a QMD that the manifest claims doesn't have one emits a warning rather than failing the check. A FAIL would have been hard to ignore; warnings get scrolled past.

The combination of (1) and (2) is what allowed 10 chapters to drift over time.

### Step 1 — Add Ruby to the CI environment

```yaml
# .github/workflows/something.yml
- uses: ruby/setup-ruby@v1
  with:
    ruby-version: '3.2'
```

The script only needs Ruby for YAML parsing (the manifests). `ruby/setup-ruby` is the standard action.

### Step 2 — Add a structure-check job

Either a dedicated workflow or a job in an existing one. Loops 1..12 plus the appendix manifests:

```yaml
- name: Structure checks
  run: |
    for d in 1 2 3 4 5 6 7 8 9 10 11 12; do
      ./scripts/check-structure.sh \"\$d\"
    done
```

The script already exits non-zero on FAIL, so this works as a CI gate as soon as the WARN → FAIL flip lands.

### Step 3 — Promote download-button drift to FAIL

In `scripts/check-structure.sh:338-358`, change the inverse mismatch path from `warn` + `total_warn++` to `fail` + `total_fail++`. Audit the other WARN paths in the script at the same time — heading drift was also WARN-only until #160 fixed Day 12's heading manifest in the same PR rather than as separate debt.

## Acceptance

- [ ] Pushing a branch with manifest drift fails CI before merge.
- [ ] No new local-only step is required of contributors — CI catches it.
- [ ] The script's other warning paths are reviewed: each is either justified (then leave as WARN with a comment explaining why) or also promoted to FAIL.

## Out of scope

- The structure check is fundamentally Quarto-specific. Tools like `quarto check` won't replace it.
- Rewriting the script in a Ruby-free language to avoid the CI dependency. Ruby is fine; the only concern is wiring the action in.

## Source

Surfaced by claude[bot] review on #273: https://github.com/lddurbin/twelve_days_to_deming/pull/273#issuecomment-$(gh api repos/lddurbin/twelve_days_to_deming/issues/273/comments --jq '.[0].id')

lddurbin:
## Summary

Closes #274. Three changes that together make manifest drift impossible to merge silently:

1. **New `.github/workflows/structure-check.yml`** — installs Ruby via `ruby/setup-ruby` (SHA-pinned to `v1.306.0` matching the repo's SHA-pin convention) and loops `scripts/check-structure.sh` against every day 1–12 and every `appendix-*-manifest.yml`. Triggers on PR + push to main, mirroring `accessibility.yml`.
2. **Download-button drift WARN → FAIL** in `scripts/check-structure.sh` — the original trigger for the issue.
3. **Heading drift WARN → FAIL** by the same logic, plus removal of dead WARN infrastructure now that no warning paths remain.

## Why both WARN paths got promoted

The original issue called out the download-button WARN explicitly and asked "audit the other WARN paths in the script." There was one other: heading drift at the line that fired when a chapter had `## ` headings that the manifest didn't list. Keeping it WARN would have left exactly the asymmetry that bit Day 12 in #160 — fail loudly when a manifest heading was absent from source, but only whisper when source headings were absent from the manifest. Promoting both keeps the rule symmetric: **manifest and source must agree**, full stop.

With both WARN call-sites gone, the supporting infrastructure (`warn()` helper, `total_warn` accumulator, `YELLOW` colour, the warnings line in the summary) is dead. Removed for clarity — easier to add back when a real WARN-level need surfaces than to keep a misleading three-state model alive in the meantime.

## Behaviour change for contributors

Before: drift accumulated silently; `WARN` lines in local runs got scrolled past.
After: pushing manifest drift fails CI before merge. No new local-only step required of contributors.

## CI loop strategy

The job accumulates failures across all 14+ manifests rather than fast-failing on the first broken day. CI minutes are cheap; getting one report that surfaces every drifted day at once beats discovering them across multiple PR pushes.

## Verification

Ran the modified script locally against every day 1–12 plus `appendix-contributions-balaji-reddie` after the WARN→FAIL flip — all pass with zero failures. So this PR turns on the gate without surfacing any latent drift that needs separate cleanup.

## Test plan

- [x] `ruby -ryaml -e 'YAML.safe_load(File.read(ARGV[0]))' .github/workflows/structure-check.yml` confirms the workflow YAML parses
- [x] `./scripts/check-structure.sh N` for N in 1..12 — all green
- [x] `./scripts/check-structure.sh --appendix contributions-balaji-reddie` — green
- [x] Exit code is 0 on a passing day (preserves existing CI semantics)
- [ ] CI run on this PR (will validate Ruby-on-Ubuntu setup + the workflow itself)

### DIFF
--- a/.github/workflows/structure-check.yml
+++ b/.github/workflows/structure-check.yml
@@ -6,6 +6,16 @@ permissions:
 on:
   pull_request:
     types: [opened, synchronize]
+    # The structure check only validates QMDs against manifests, plus the
+    # images they reference and the script/workflow themselves. Narrow the
+    # trigger to those paths so unrelated PRs (CSS-only, JS-only, prose
+    # outside content/, etc.) don't burn CI minutes on it.
+    paths:
+      - 'content/**'
+      - 'workflow/validation/**'
+      - 'assets/images/**'
+      - 'scripts/check-structure.sh'
+      - '.github/workflows/structure-check.yml'
   push:
     branches:
       - main
--- a/.github/workflows/structure-check.yml
+++ b/.github/workflows/structure-check.yml
@@ -1,0 +1,47 @@
+name: Structure Check
+
+permissions:
+  contents: read
+
+on:
+  pull_request:
+    types: [opened, synchronize]
+  push:
+    branches:
+      - main
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  structure:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
+
+    - name: Setup Ruby
+      uses: ruby/setup-ruby@c4e5b1316158f92e3d49443a9d58b31d25ac0f8f  # v1.306.0
+      with:
+        ruby-version: '3.2'
+
+    - name: Run structure checks
+      run: |
+        overall_fail=0
+        for d in $(seq 1 12); do
+          echo "::group::Day $d"
+          ./scripts/check-structure.sh "$d" || overall_fail=1
+          echo "::endgroup::"
+        done
+
+        shopt -s nullglob
+        for manifest in workflow/validation/appendix-*-manifest.yml; do
+          slug=$(basename "$manifest" | sed -E 's/^appendix-(.+)-manifest\.yml$/\1/')
+          echo "::group::Appendix $slug"
+          ./scripts/check-structure.sh --appendix "$slug" || overall_fail=1
+          echo "::endgroup::"
+        done
+
+        exit "$overall_fail"
--- a/scripts/check-structure.sh
+++ b/scripts/check-structure.sh
@@ -12,9 +12,8 @@
 # download button. Appendix manifests can opt out of interactive checks for
 # prose-only content (set `interactive_checks: false`).
 #
-# Requires: ruby (for YAML parsing — ships with macOS)
-# Note: CI (Ubuntu) does not install Ruby by default. If wiring into CI,
-# add ruby to the workflow's setup steps.
+# Requires: ruby (for YAML parsing — ships with macOS).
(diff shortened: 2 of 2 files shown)


lddurbin:
Thanks for the review.

**(1) Path filter — implemented in c9dd8f3, with a slight broadening.**

Your exact list (`content/days/**`, `workflow/validation/**`, the script + workflow) would have missed two paths the script actually reads:
- `content/appendix/**` (appendix manifests can declare `content_dir`, so the check covers appendix QMDs too)
- `assets/images/**` (the script verifies that every figure path in a manifest exists on disk, and that every `` in a QMD points to a real file)

A PR that deleted an image or moved one to a different filename without touching its QMD would silently skip the check under the narrower filter, and we'd lose a real failure mode. Broadened to `content/**` + `assets/images/**` to cover those, plus a brief comment on the trigger explaining the rule so future maintainers can keep it in sync if the script grows.

**(2) Ruby version pin — leaving as `3.2`.**

You explicitly flagged this as "very low risk" and "worth considering if you ever need to debug" — and I think it's the right kind of suggestion to leave on the table without acting on. The script uses only stdlib `YAML.safe_load`, whose behaviour is rock-stable across 3.2.x. Pinning to `3.2.7` (or any specific patch) would buy determinism we don't need and add a manual-bump cadence whenever the runner image's default 3.2.x ages out. I'd rather take the floating patch and pin if a real reproducibility issue ever surfaces.

Open to revisiting if you've seen a concrete case where 3.2.x patch differences bit you.

PR #307, opened 2026-05-02
CSP: self-host Source Sans Pro + Lato woff2 (#302)

### LINKED ISSUE #302: Decide font hosting strategy: allow Google Fonts vs self-host woff2
## Context

Issue #299 added `data:` to `font-src` in `_quarto.yml`'s CSP `<meta>` tag (PR-pending), which fixes the anchorjs heading-anchor glyph regression — the headline accessibility win.

With that landed, the only remaining `font-src`/`style-src` CSP blocks are for Cosmo's Google Fonts requests:

- **Source Sans Pro** (300/400/700) — ~14 woff2 files from `fonts.gstatic.com`
- **Lato** (400/700) — ~6 woff2 files from `fonts.gstatic.com`
- The Google Fonts CSS stylesheet from `fonts.googleapis.com`

Current behaviour: pages fall back to the OS system-font stack. Not broken — just not Cosmo's intended typography.

## Decision needed

Two paths; both fix the typography but trade differently. (Option (a) — leave as-is — was already chosen and shipped in #299.)

### Path B — allow Google Fonts in the CSP

```yaml
# _quarto.yml (header-includes meta)
font-src 'self' data: https://fonts.gstatic.com;
style-src 'self' 'unsafe-inline' https://fonts.googleapis.com;
```

- **Effort**: 1 line of YAML.
- **Privacy cost**: every page load reveals user IPs to Google's font servers.
- **Offline**: fonts fail → system-font fallback (today's behaviour).

### Path C — self-host woff2 files

1. Download Source Sans Pro (300/400/700) and Lato (400/700) woff2 files from Google Fonts.
2. Place them under `assets/fonts/`.
3. Override Cosmo's `@font-face` declarations in `assets/styles/main.css` to point to the local copies.
4. Keep CSP at `font-src 'self' data:` — no third-party allowance.

- **Effort**: ~30 min, plus ~9 small woff2 commits (≈50 KB each).
- **Privacy**: best — zero third-party requests on page load.
- **Offline**: fonts load anywhere the static site does.
- **Maintenance**: woff2 files are effectively static; no ongoing updates.

## Recommendation

**Path C.** The project already self-hosts almost everything and ships a CSP that explicitly minimises third-party connections (`default-src 'self'` plus a small allow-list). Adding `fonts.googleapis.com` + `fonts.gstatic.com` would be the largest single privacy concession the project has made — for a purely cosmetic gain. Self-hosting matches the existing privacy posture.

## Acceptance criteria

- [ ] Source Sans Pro (300/400/700) and Lato (400/700) render in body and headings (compare to Cosmo's intended look on a fresh chapter page).
- [ ] No `Content-Security-Policy: … blocked … (font-src)` warning in browser console on a freshly-loaded chapter page.
- [ ] No `Content-Security-Policy: … blocked … (style-src)` warning on a freshly-loaded chapter page.
- [ ] If Path C is chosen: woff2 files committed under `assets/fonts/` and `@font-face` declarations live in `assets/styles/main.css`.

## Out of scope

- Re-litigating the broader CSP. Only `font-src` and possibly `style-src` need touching.
- Anchorjs heading-glyph fix — already shipped in #299's option (a).

## References

- Predecessor: #299 (option (a) shipped — anchorjs `data:` font now allowed).
- CSP definition: `_quarto.yml:191`.
- Surfaced during: console-noise audit while sweeping right-floating image layouts (May 2026).

lddurbin:
## Summary
- Self-host Source Sans Pro (300/400/700) and Lato (400/700/400i) — Path C from #302.
- Six woff2 files (latin subset only, ~88 KB total) committed under `assets/fonts/`.
- Matching `@font-face` declarations appended to the SELF-HOSTED FONTS section in `assets/styles/main.css`.
- New `assets/styles/disable-google-fonts.scss` sets `$web-font-path: false`, which short-circuits the Bootswatch `@import` so the compiled bootstrap bundle no longer references `fonts.googleapis.com` / `fonts.gstatic.com`.
- `_quarto.yml` `theme:` arrays updated to layer that SCSS partial onto cosmo + darkly.

Closes #302.

## Why Path C (vs Path B allow-list)
Path B would have widened CSP to allow `https://fonts.gstatic.com` (font-src) and `https://fonts.googleapis.com` (style-src) — every reader's IP would be revealed to Google on every page load for purely cosmetic gain. Path C matches the project's existing posture (DM Serif Display + OpenDyslexic are already self-hosted; CSP otherwise stays at `default-src 'self'` plus a tiny allow-list).

## Latin subset only
The single non-latin char in body content is "ū" (one Japanese book transliteration in `content/appendix/11-references-and-sources.qmd:40`). It falls through to the system `sans-serif` family — visually inconsistent on that one word but no broader regression. If Maori macron content arrives later, latin-ext can be added in a follow-up (~80 KB).

## Test plan
- [x] Static: rebuilt bootstrap bundles contain **zero** references to `fonts.googleapis.com` / `fonts.gstatic.com` (bundle hash changed `5ccb035bea` → `14858d2fc`, confirming Sass re-compile applied the override).
- [x] Static: bundles still emit `font-family: "Source Sans Pro", ...` (light) and `font-family: Lato, ...` (dark) → browser will resolve those names against our local `@font-face`.
- [x] Static: woff2 files all start with `wOF2` magic bytes (valid).
- [x] Static: CSP at `_quarto.yml:191` unchanged — `font-src 'self' data:; style-src 'self' 'unsafe-inline'`.
- [ ] Visual: open a chapter page in light + dark theme on `quarto preview`; confirm body text renders in Source Sans Pro / Lato (not the system fallback) and there are zero `Content-Security-Policy: … blocked` warnings in the console. _**Reviewer: visual confirmation needs a browser — I couldn't run that part of the check from CLI.**_

## Notes for reviewer
- No new external requests. CSP allow-list is unchanged.
- OFL 1.1 covers redistribution of both fonts as-is. Project precedent (DM Serif Display) is to ship the woff2 files without a per-font licence text; this PR follows that pattern.


lddurbin:
Thanks for the careful review. Acted on the second concern; pushing back on the first.

### `unicode-range` descriptors — added in 24dadbc

Agreed on the practice, but for a different reason than the one given. Browsers without `unicode-range` don't render tofu for missing glyphs — they fall through the `font-family` cascade. Cosmo's stack ends in `sans-serif`, so `ū` at `content/appendix/11-references-and-sources.qmd:40` would have rendered cleanly in the system font even without the descriptor.

That said, declaring `unicode-range` is the canonical way to represent a subset font, it's what Google's served CSS literally includes for these faces, and it's basically free. So I added it — using Google's exact emitted range (which differs slightly from the one in your comment: it includes combining macron/diaeresis/vertical-line-below `U+0304/0308/0329` and omits superscript four `U+2074`).

### Italic for Source Sans Pro — declining

The premise that Cosmo "requests italic cuts alongside normal weights" doesn't match the URL Cosmo actually emits:

```
fonts.googleapis.com/css2?family=Source+Sans+Pro:wght@300;400;700&display=swap
```

Path is `wght@300;400;700` — weights only, no `ital,` prefix. Google would have served three faces, all `font-style: normal`, and any `<em>` in light-theme content was synthesized from the upright cut **before this PR too**. Self-hosting matches Cosmo's specification byte-for-byte; adding an SSP italic would *change* the rendering versus what light-theme readers were seeing yesterday.

Compare to the Lato URL, which **does** request italic (`ital,wght@0,400;0,700;1,400`) — and accordingly we self-host `lato-400-italic.woff2`. The two families diverge here because Cosmo's authors made different choices for the two themes; this PR preserves both.

If italic-quality on light-theme prose becomes a real concern (e.g. Neave's `*Quality, Productivity, and Competitive Position*` in body text doesn't look right), that's worth a separate issue framed as "Cosmo's font selection isn't great for our prose" rather than a self-hosting follow-up.

PR #319, opened 2026-05-23
Site: record Neave's public-release blessing + enable sitemap/footer

lddurbin:
## Summary

Dr Henry Neave reviewed the live site in May 2026 and explicitly authorised it to be made publicly available. This PR records the endorsement in user-facing copy and shifts the site's posture from *private with permission to transcribe* to *publicly endorsed by the author*.

- **`about-this-edition.qmd`** — new `## Dr Neave's blessing` section, placed between *Why this edition exists* and *What has changed, and what hasn't*. Quotes Neave directly (*"do make it as publicly available as possible…"*) and explains what the endorsement covers: faithful transcription and the accessibility framing.
- **`README.md`** — drops the "(Work in Progress)" tag on the Live Course entry and replaces the generic `## 📄 License` blurb with a specific `## 📄 License and Permissions` section that credits Neave, points upstream to the NZOQ PDFs, and asks for citation when sharing.
- **`_quarto.yml`** — sets `book.site-url: https://deming.leedurbin.co.nz`, which causes Quarto to auto-generate `sitemap.xml` (with absolute URLs and lastmod timestamps) and `robots.txt` (no `Disallow` directives → fully indexable). Adds a `page-footer` with left/right slots: course content attribution to Neave on the left, web-edition credit + GitHub source link on the right.

## Why this matters

Until now the site has been deliberately low-profile pending Neave's approval. With his blessing in hand, the discoverability posture should match: search engines need a sitemap to index efficiently, and every page should carry persistent attribution so the provenance is unambiguous to readers and re-publishers.

## Test plan

- [x] `quarto render` completes clean across all 121 files
- [x] `_book/sitemap.xml` generated with absolute URLs + lastmod timestamps
- [x] `_book/robots.txt` generated with sitemap reference and no Disallow rules
- [x] Footer renders on `about-this-edition.html` with `nav-footer-left` / `nav-footer-right` classes (theme-aware styling via existing CSS)
- [x] `ruby -ryaml` validates `_quarto.yml` syntax
- [ ] Visual spot-check after deploy: footer layout on mobile, dark mode footer contrast, sitemap accessible at `/sitemap.xml`

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,10 +59,10 @@ Each day includes:
 
 ## 📥 Accessing the Course Material
 
-### Interactive Online Version (Work in Progress)
+### Interactive Online Version
 - **Live Course**: [deming.leedurbin.co.nz](https://deming.leedurbin.co.nz) - The interactive version with embedded activities and exercises
 
-**Current Status**: All 12 days are now fully converted from the original PDF format into the interactive experience. Content refinements, accessibility improvements, and activity polish are ongoing.
+**Current Status**: All 12 days are fully converted from the original PDF format into the interactive experience. Dr Henry Neave reviewed this edition and, in May 2026, gave his blessing for it to be made publicly available. Accessibility improvements and activity refinements are ongoing.
 
 **Enhanced Features in the Online Version:**
 - **Interactive Clock Indicators**: Built-in timing guidance to help you pace your study sessions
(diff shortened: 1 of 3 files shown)


lddurbin:
Thanks for catching the `·` separator — agreed it's worth fixing given the site's accessibility framing (made even more explicit in the new *Dr Neave's blessing* section in this same PR).

Pushed `61b0a73` (will update once pushed) wrapping it in `<span aria-hidden="true">·</span>`. Went with the inline-HTML option over a shortcode since Quarto's `page-footer` accepts markdown-with-HTML directly — no need to introduce a new filter for a one-off decorative character. The two adjacent `<a>` elements now get announced as discrete links via their link role, which is the AT parsing cue we want.

On the trailing-newline-from-`|`-block-scalars note: confirmed harmless — Quarto wraps the slot content in a `<p>` and trims appropriately. No action there.

PR #337, opened 2026-05-23
CI: nightly external link-rot watch with lychee (#284)

### LINKED ISSUE #284: Nightly external link-rot watch (lychee or wcag-link-checker)
## Context

v0.1.0 ships **144 internal image refs and ~250 external links** (concentrated in `content/appendix/11-references-and-sources.qmd` and inline citations across all 12 days). External link rot is the silent failure mode for educational content: Wayback Machine links shift, deming.org.uk has reorganised in the past (closed issue #239 cleaned up dead links once already), GQC and NZOQ host content move under their own redesigns.

The v0.1.0 baseline evaluation (`docs/evaluations/v0.1.0-baseline-2026-04-27.md` §8 Tier 3) recommends a nightly link-check job to catch external link rot before users do.

## Scope

1. Add a scheduled GitHub Actions workflow (e.g. `.github/workflows/link-check.yml`) running on a nightly cron (suggest 06:00 UTC = 18:00 NZT, off-peak).
2. Use a tool like [lychee](https://github.com/lycheeverse/lychee) (Rust, fast, well-maintained) or [linkchecker](https://github.com/linkchecker/linkchecker). Lychee has a published GitHub Action and good `--cache` support.
3. Configure to:
   - Crawl the rendered `_book/` output (or run against checked-out QMDs with markdown link extraction — the former is more accurate as it includes auto-generated cross-refs).
   - Skip internal anchors (already covered by structure-check / cross-ref Python audit).
   - Treat 404, 410, and connection-refused as failures; treat 429 (rate-limited) as a soft warning to avoid noisy false positives.
   - Emit a markdown summary into the workflow run artifacts.
4. Failure mode: open a GitHub issue (or post to a single rolling tracking issue) listing the broken links, rather than failing the whole workflow loudly. Nightly rotnotices are advisory, not gating.
5. Exclude lists: any links we know are flaky but valid (e.g. some academic mirrors). Keep exclusions in `.lycheeignore` so the rationale is visible.

## Acceptance criteria

- [ ] Workflow runs on schedule and on-demand via `workflow_dispatch`.
- [ ] Workflow does not run on every PR (cost and noise).
- [ ] First run produces a baseline; subsequent runs only flag deltas.
- [ ] Failure path opens or updates a tracking issue rather than blocking merges.
- [ ] `.lycheeignore` (or equivalent) is documented.

## References

- Evaluation §8 Tier 3, recommendation 9.
- Prior dead-link cleanup: closed issue #239 (deming.org.uk, rqoq.org.nz).
- External link concentration: `content/appendix/11-references-and-sources.qmd`.

lddurbin:
## Summary
- Adds `.github/workflows/link-check.yml` — a scheduled (06:00 UTC / 18:00 NZT nightly) + `workflow_dispatch` job that runs `lycheeverse/lychee-action` against every `.qmd` in the repo and rolls findings into a single tracking issue rather than failing CI.
- Adds `.lycheeignore` as an empty stub with the rationale for when an exclusion is appropriate.
- Closes #284.

## Design notes
- **External-only scope.** The workflow filters with `--scheme http --scheme https`, so internal cross-references (`../day-NN/file.qmd#sec-pageN` etc.) are deliberately skipped — those are covered by `structure-check` and `interday-audit`.
- **Source-not-render.** Scanning `.qmd` source directly skips the renv-restore + Quarto render path; external URLs are written verbatim in source, so this is accurate enough for rot detection while keeping a nightly tick cheap (no R setup, no system deps).
- **Accept policy.** `200..=299` (success), `403` (CDN-fronted hosts that 403 a bare GET but render in a browser), `429` (rate-limited; advisory only). Anything else is flagged.
- **Single rolling issue.** On findings, the workflow searches for an open issue titled `[link-rot] Nightly external link check report`. If it exists, the night's report is appended as a comment. If not, a new issue is opened with the `maintainability`, `github_actions`, and `severity: low` labels (mirroring #284's own).
- **Cache.** The `.lycheecache` file is persisted across runs via `actions/cache`; combined with `--max-cache-age 1d`, unchanged URLs are not re-probed every night.
- **Failure containment.** `fail: false` + `failIfEmpty: false` + a `[ -s lychee-report.md ]` guard mean a network blip or CLI parse error surfaces as a clear note in the tracking issue rather than a red workflow with no signal.
- **PR triggers omitted.** The workflow does not run on PRs (cost + noise per the issue's AC).
- **Action SHAs.** All three external actions (`actions/checkout`, `actions/cache`, `lycheeverse/lychee-action`, `actions/upload-artifact`) are pinned to commit SHAs matching the repo's existing convention.

## Acceptance criteria from #284
- [x] Workflow runs on schedule and on-demand via `workflow_dispatch`.
- [x] Workflow does not run on every PR.
- [x] First run produces a baseline; subsequent runs only flag deltas (via lychee cache + `--max-cache-age 1d`).
- [x] Failure path opens or updates a tracking issue rather than blocking merges.
- [x] `.lycheeignore` is documented.

## Test plan
- [ ] Once merged, manually trigger via Actions → Link Check (Nightly) → Run workflow, and confirm:
  - [ ] A tracking issue is opened iff there are real broken links.
  - [ ] If the tracking issue already exists, a fresh dated comment is appended.
  - [ ] The `lychee-report` artifact is attached to the run.
- [ ] Confirm the cron fires on the next 06:00 UTC tick (visible in the Actions runs list).

### DIFF
--- a/.github/workflows/link-check.yml
+++ b/.github/workflows/link-check.yml
@@ -1,4 +1,4 @@
-name: Link Check (Nightly)
+name: Link Check (Weekly)
 
 permissions:
   contents: read
@@ -6,8 +6,11 @@ permissions:
 
 on:
   schedule:
-    # 06:00 UTC daily = 18:00 NZT, off-peak for both UTC and NZ audiences.
-    - cron: '0 6 * * *'
+    # Monday 06:00 UTC = Monday 18:00 NZT, start of NZ work week so findings
+    # land just as the week begins. Weekly cadence (rather than nightly) is
+    # tuned to the corpus: ~53 external URLs targeting slow-moving educational
+    # / academic sites, where rot happens on a timescale of weeks not hours.
+    - cron: '0 6 * * 1'
   workflow_dispatch:
 
 concurrency:
@@ -22,7 +25,7 @@ jobs:
     - name: Checkout repository
       uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
 
-    # Persist lychee's per-URL cache across runs so each nightly tick only
+    # Persist lychee's per-URL cache across runs so each weekly tick only
     # re-probes URLs whose entries are older than --max-cache-age, plus any
     # URL that wasn't in the cache. The write key uses run_id so every run
     # produces a fresh cache entry; the restore-keys prefix picks up the
@@ -49,7 +52,7 @@ jobs:
         #     and 429 (rate-limited; advisory only, not real rot)
         args: >-
           --cache
-          --max-cache-age 1d
+          --max-cache-age 7d
           --no-progress
           --include-fragments false
           --scheme http
@@ -75,8 +78,8 @@ jobs:
         path: lychee-report.md
         retention-days: 30
 
-    # Roll all nightly reports into a single tracking issue rather than
-    # opening a new one per night. Match by title sentinel "[link-rot]"
+    # Roll all weekly reports into a single tracking issue rather than
+    # opening a new one per week. Match by title sentinel "[link-rot]"
     # so the lookup is deterministic and human-readable.
     - name: Update tracking issue
       if: steps.lychee.outputs.exit_code != '0'
@@ -87,7 +90,7 @@ jobs:
       run: |
         set -euo pipefail
 
-        TITLE="[link-rot] Nightly external link check report"
+        TITLE="[link-rot] Weekly external link check report"
 
         # Prepend run metadata so each comment / issue body is self-dating.
         # Guard the cat: if lychee bailed before writing the report (e.g. CLI
--- a/.github/workflows/link-check.yml
+++ b/.github/workflows/link-check.yml
@@ -1,0 +1,126 @@
+name: Link Check (Nightly)
+
+permissions:
+  contents: read
+  issues: write
+
+on:
+  schedule:
+    # 06:00 UTC daily = 18:00 NZT, off-peak for both UTC and NZ audiences.
+    - cron: '0 6 * * *'
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}
+  cancel-in-progress: false
+
+jobs:
+  link-check:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
+
+    # Persist lychee's per-URL cache across runs so each nightly tick only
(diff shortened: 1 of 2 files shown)


lddurbin:
Thanks for the careful review. Validated both points; pushed one fix, pushing back on the other.

### 1. `--include-fragments false` — fixed (2f129f8)

You're directionally correct but the actual mechanism is worse than "stray positional arg." In lychee v0.24.2 (bundled with `lychee-action@v2.8.0`), `--include-fragments` is **not** a boolean — it's `Option<FragmentMode>` defined with `require_equals = true` and `default_missing_value = "anchor-only"` ([source](https://github.com/lycheeverse/lychee/blob/lychee-v0.24.2/lychee-bin/src/config/mod.rs)):

```rust
#[arg(
    long,
    default_missing_value = "anchor-only",
    num_args = 0..=1,
    require_equals = true,
    value_name = "none|anchor-only|text-only|full",
    ...
)]
include_fragments: Option<FragmentMode>,
```

So `--include-fragments false` parses as:
1. Bare `--include-fragments` (no `=`) → `default_missing_value` kicks in → fragment checking gets **enabled** in `anchor-only` mode (the opposite of intent), and
2. `false` becomes a positional file/URL arg that lychee then errors on.

Default behavior (`include_fragments: None`) already disables fragment checking, so dropping the line restores the intended semantics. Fix is one line removed.

### 2. `FORCE_JAVASCRIPT_ACTIONS_TO_NODE24` — declining

Checked the repo's pattern more carefully and I think the case for adding it here is weaker than the comment suggests:

- The env block is **not** uniformly applied across workflows. `structure-check.yml` (Ruby + bash, no r-lib) omits it.
- The rationale on `accessibility.yml:17-18` is explicit: _"Opt r-lib/actions into Node 24 ahead of the Sept 2026 Node 20 removal. Remove once r-lib publishes a v3 tag that runs on Node 24 natively."_ The pattern is **r-lib/actions-specific**, not a blanket "set this on every workflow."
- All four JS-touching actions in `link-check.yml` are already on Node 24 (or composite):
  - `actions/checkout@v6.0.2` — Node 24 native
  - `actions/cache@v5.0.5` — Node 24 native
  - `lycheeverse/lychee-action@v2.8.0` — composite action, shells out to a downloaded lychee binary
  - `actions/upload-artifact@v5.0.0` — Node 24 native (v5 was the Node 24 release)

So adding the env block would either (a) require a new justification comment alongside the structurally-different r-lib one, or (b) copy the r-lib comment in a context where it's misleading. Neither feels like a net gain. Happy to revisit if the repo's pattern shifts toward "set everywhere" — for now I'm keeping it consistent with `structure-check.yml`'s precedent.

Issue #316, opened 2026-05-13
Day 12 radar diagrams


lddurbin:
Tracking: #309

Depends on the foundation issue.

## Scope

| File | qmd |
|---|---|
| `assets/images/day-12/radar-diagram-large.png` | Day 12.02 |
| `assets/images/day-12/radar-diagram-small.png` | Day 12.02 |
| `assets/images/day-12/radar-diagram-key-table.png` | Day 12.02 — separate effort to render via `gt`; out of scope |

## Implementation notes

- Radar charts in ggplot2 are awkward — recommend `ggradar` package or a custom polar-coord implementation. Decide in the issue's design phase before implementing.
- The radar diagram is also a *learner activity* (learners fill it in for themselves). The "small" version is likely a printable template. Keep that one as PNG; regenerate only the "large" reference version with annotations.

## Acceptance criteria

- [ ] Radar reference diagram regenerated.
- [ ] Decision documented on charting package vs. hand-rolled.
- [ ] Deviations log updated.

lddurbin:
Decomposed into sub-issues:

- [x] #360 — Implement radar-diagram-large.png (with ggradar-vs-polar spike) (S)
- [x] #361 — Keep small + key-table as PNG (deviations log) (XS)

lddurbin:
Closing — both sub-issues merged and all three acceptance criteria met.

**Sub-issues:**
- #360 — radar-diagram-large.png replaced with R/ggplot2 (PR #381) ✓
- #361 — radar-diagram-small.png + key-table kept as PNG, documented in deviations log (PR #381) ✓

**Acceptance criteria:**
- Radar reference diagram regenerated ✓ (hand-rolled `coord_fixed()` helper `radar_diagram_plot()`, added after `specification_limits_diagram()` in `R/functions/main-functions.R`)
- Decision documented on charting package vs hand-rolled ✓ (`ggradar` rejected after 30-min spike — Neave's radar is a Cartesian crosshair with four orthogonal arms and a 0-to-5 linear scale, not a closed-polygon radar; details in PR #381's deviations entry)
- Deviations log updated ✓

Orphan-PNG cleanup for the large radar tracked under the epic-#309 policy (separate follow-up PR after production verification). The small + key-table PNGs are kept-as-PNG by design, not orphans.

Issue #311, opened 2026-05-13
Day 3 single-panel charts: run charts, individual control charts, Taguchi loss


lddurbin:
Tracking: #309

Depends on the foundation issue (dark-mode-safe theme + conventions).

## Scope

Replace with R-generated figures:

| File | qmd | Notes |
|---|---|---|
| `assets/images/day-03/run-chart-first-process.png` | [content/days/day-03/03-the-importance-of-time.qmd](content/days/day-03/03-the-importance-of-time.qmd) | reuse `run_chart_plot()` |
| `assets/images/day-03/favourite-example-run-chart.png` | [content/days/day-03/07-a-favourite-example.qmd](content/days/day-03/07-a-favourite-example.qmd) | proportion-defective data |
| `assets/images/day-03/favourite-example-control-chart.png` | [content/days/day-03/07-a-favourite-example.qmd](content/days/day-03/07-a-favourite-example.qmd) | proportion-defective control chart |
| `assets/images/day-03/sales-control-chart.png` | [content/days/day-03/04-more-on-the-sales-data.qmd](content/days/day-03/04-more-on-the-sales-data.qmd) | already has an R version; delete orphan PNG |
| `assets/images/day-03/technical-aid-8-control-chart.png` | Day 3.08 | individual-X with UCL/LCL |
| `assets/images/day-07/taguchi-loss-function.png` | Day 7.06 | parabola via `stat_function()` |
| `assets/images/day-07/specification-limits.png` | Day 7 | step / shaded region |

## Acceptance criteria

- [ ] Each chart above is rendered from R inline in its `.qmd`.
- [ ] PNGs deleted from `assets/images/`.
- [ ] Captions and alt text preserved (see deviations log entry).
- [ ] Dark mode verified.
- [ ] Deviations log updated.

## Implementation notes

- Source data: transcribe values from Neave into a small CSV under a new `R/data/` directory or as inline `c(...)` vectors in the `.qmd`. Prefer CSV when the data is shared across chapters.
- Taguchi loss function: `y = k(x - target)^2` — needs a chosen `k` to match the visual proportions of Neave's figure; document choice in deviations log.

Issue #309, opened 2026-05-13
Epic: Replace cropped chart PNGs with R-generated figures


lddurbin:
## Background

Currently the book draws most of its charts as PNGs cropped from Neave's source PDF and stored under [assets/images/](assets/images/). The cropping workflow is documented in [CLAUDE.md](CLAUDE.md). Inspired by Kieran Healy's piece [*Using Quarto to write a book*](https://kieranhealy.org/blog/archives/2026/03/09/using-quarto-to-write-a-book/), this epic proposes regenerating charts whose underlying data is reproducible (run charts, control charts, histograms, distribution illustrations, funnel-experiment visuals, radar diagrams) directly in R/ggplot2 — while leaving photos, iconic diagrams, and historical artefacts as cropped images.

The pattern is already proven for Day 2 and parts of Day 3 ([R/functions/main-functions.R](R/functions/main-functions.R)). This epic extends it.

## Why

- **Sharpness:** SVG output is crisp at any zoom, vs. 2550×3300 PNGs scaled down.
- **Dark mode:** Cropped PNGs can't recolour; ggplot output can be themed via [main.css](assets/styles/main.css) variables.
- **Accessibility:** Text and lines in SVG are screen-reader inspectable in ways bitmap alt text is not.
- **Reproducible-research ethos:** The same datasets the book *teaches* sit in the repo, runnable.
- **Maintainability:** Tweaks (typos in labels, axis adjustments) become code edits, not new crop passes.

## Policy: what we convert vs. what we keep

**Convert (R-generated):**
- Run charts and individual-X control charts of Neave's tabulated data
- Histograms of Neave's tabulated data (Ford shafts, two-histograms-24-values)
- Funnel experiment "track" dot-plots and dice-sequence visualisations
- Distribution and central-limit illustrations in the Optional Extras
- Taguchi loss-function parabola and specification-limits diagram (Day 7)
- Radar diagrams (Day 12)

**Keep as cropped images (out of scope):**
- Photographs (Mack McIngvale, Deming portraits, calipers, the red-bead paddle)
- Diagrams that are *concept maps*, not data: Joiner Triangle, Shewhart circle, Scholtes SoPK diagram, organisation charts, flow diagrams, theory-of-knowledge circle
- Handwritten artefacts from Neave's seminars (Day 2 `image-2.png` through `image-14.png` — these are historical seminar results, the handwritten quality *is* the content)
[...]

PR #339, opened 2026-05-25
R figures: dark-mode-safe ggplot foundation (#310)

### LINKED ISSUE #310: Foundation: dark-mode-safe ggplot theme + R-figure conventions
Tracking: #309

## Background

Before adding more R-generated charts, the foundation needs a dark-mode-safe theme. `run_chart_theme()` in [R/functions/main-functions.R](R/functions/main-functions.R) currently hard-codes light colours (`grey80`, `black`, `#cccccc`), so the existing Day 2 / Day 3 R charts are visually broken under the `darkly` theme. Every subsequent chart issue would compound this regression.

## Scope

1. Refactor `run_chart_theme()` so colours come from constants that can be swapped per theme.
2. Decide the dark-mode delivery mechanism. Two options:
   - **(a) Dual rendering:** generate two SVGs per chart (`-light.svg`, `-dark.svg`) and toggle via CSS `prefers-color-scheme` / theme attribute. Pro: full chroma freedom. Con: doubles render time, doubles asset count.
   - **(b) CSS-only restyle:** emit SVG with named CSS classes/`currentColor` for axes/text, restyle in [main.css](assets/styles/main.css). Pro: single asset, fast. Con: ggplot needs an `svglite` device and post-processing to inject classes; some elements (panel.background) need transparent fill.
   - Recommendation: **(b) for axes/text/gridlines, (a) only if a chart genuinely needs different ink colours**.
3. Add an R-figure conventions doc-comment block at the top of `R/functions/main-functions.R` covering: default figure dimensions, when to use `fig-cap` vs caption-in-image, alt-text conventions (Quarto drops the alt attr from plain `cap`; use `{.lightbox}` to keep both caption and alt), naming of helper functions.
4. Add a `histogram_plot()` helper (counts-based, optional bin labels, frequency or relative-frequency Y-axis). Needed by the Day 3 histograms and Optional Extras issues downstream.
5. Verify existing Day 2 and Day 3 R charts still match their captions after the theme refactor.

## Acceptance criteria

- [ ] Existing R charts render correctly in both light and dark modes (visual diff check).
- [ ] `histogram_plot()` exists with documented examples.
- [ ] Conventions doc-comment is at the top of `R/functions/main-functions.R`.
- [ ] No visible regression in Day 2 / Day 3 charts.

## Out of scope

- Converting any new figures (that's the per-day issues).
- Moving `R/functions/` into a package.

lddurbin:
## Summary

Foundation issue for the [Replace cropped chart PNGs with R-generated figures](https://github.com/lddurbin/twelve_days_to_deming/issues/309) epic. Before adding any new R chart, the existing run-chart theme needed to be dark-mode safe — every subsequent figure issue would have compounded the regression of black axis ink on Darkly's ~#222 body.

What changed:
- **Switched chart device PNG → SVG (svglite)** so the chart artifacts live as text-based vector data, with `bg = "transparent"` so the dark page colour bleeds through.
- **Refactored `run_chart_theme()`** to use named ink constants. Major gridlines at `#808080` clear WCAG 1.4.11 (3:1 contrast for essential non-text graphics) on both white (3.95:1) and Darkly (~#222 → 3.51:1) without a per-mode branch.
- **Added a CSS filter rule** on `body.quarto-dark .cell-output-display img[src$=".svg"]`: `filter: invert(1) hue-rotate(180deg)`. Invert flips brightness for axes/text; the 180° hue-rotate cancels the chromatic flip so the red data line and blue control limits stay roughly on-hue.
- **Added `histogram_plot()`** (counts / relative-frequency Y axis, optional bin labels). Downstream chart issues #311–#316 will share this helper.
- **Added an R-figure conventions doc-comment block** at the top of `R/functions/main-functions.R` covering colour tokens, default dimensions, the alt-text + lightbox quirk ([[feedback_image_alt_lightbox]]), and the dark-mode mechanism — so the next contributor (or me, in three weeks) knows the contract.

## Why CSS-filter rather than inline-SVG + element selectors

The issue's option (b) sketch envisaged "emit SVG with CSS classes / `currentColor` for axes/text". I tried that path first — wrote a `knit_hooks$set(plot = …)` override to inline svglite output rather than letting Quarto wrap it in ``. It worked in isolation but **Quarto's knitr engine re-installs its own plot hook after `.Rprofile` runs**, clobbering the override. Rather than fight Quarto's hook ordering with a per-chunk setup or a brittle post-render rewrite, I chose the CSS-filter idiom — it works on ``-wrapped SVG without reaching into the markup, keeps the artifacts as separate cacheable files, and produces visually equivalent dark mode through the invert+hue-rotate round-trip.

The conventions block documents the colour-token contract so any future contributor writing a chart helper picks ink that survives the round-trip cleanly.

## Test plan

- [x] `./scripts/check-structure.sh 2` — 59/59 pass
- [x] `./scripts/check-structure.sh 3` — 89/89 pass
- [x] Day 2 page 1, Day 2 page 4, Day 3 page 4 all re-render cleanly with the new svglite device
- [x] Built SVG verified: transparent background, `#808080` gridlines, unstyled `<text>` (no inline fill — picks up the filter)
- [ ] **Visual sign-off needed**: open `quarto preview`, toggle to dark mode on Day 2 pages 1 & 4 and Day 3 page 4, confirm:
  - Axis labels readable on dark background
  - Gridlines visible but not noisy
  - Red data line still recognisably red
  - Blue control limits still recognisably blue
  - Light mode unchanged

## Risks / things to watch

- `renv.lock` gains svglite (2.2.2) plus its transitive deps (stringi, systemfonts, textshaping) and minor patch bumps for cli/cpp11/rlang. svglite 2.1.2 (the previous lockfile-referenced version) needs to compile against libpng from source, which the binary 2.2.2 avoids — staying on 2.2.2 keeps CI lean.
- The chunks call `dpi=300`, which is meaningful for PNG and inert for SVG — kept as-is so existing call sites need no edits.

Closes #310


lddurbin:
Thanks — all three items validated and addressed in `5d1385c`. Quick rundown of what I did and the reasoning where I diverged from your suggested fix.

### 1. Missing `} /* end @media not print */` — fixed

Added the closing brace back between the SVG-filter rule and the high-contrast block (`assets/styles/main.css:1402-1404`). Brace count is balanced again (238 ↔ 238 across the file).

I kept the SVG filter *inside* `@media not print` rather than lifting it out. Your analysis nailed why this is actually the desired behaviour — a dark-mode user printing the page wants the original axis ink on white paper, not the inverted version on whatever the printer makes of `#222`. So I added an inline comment on the rule documenting that the print-exclusion is load-bearing rather than accidental:

```css
body.quarto-dark .cell-output-display img[src$=".svg"] {
  /* Sits inside the surrounding `@media not print` so transparent-bg
     SVGs print without the inversion (original axis ink on white paper
     is what we want for hard copy). */
  filter: invert(1) hue-rotate(180deg);
}
```

### 2. Orphan comment in `R/setup.R` — fixed, slightly more aggressively

You suggested deleting lines 24–29 and re-wording line 13 (`inline SVG` → `external SVG files`). I went a step further and folded the surviving `Note:` paragraph into the same comment block, so there's now one paragraph describing the whole emit-SVG-then-CSS-filter mechanism instead of two paragraphs separated by a deleted middle block. Net result is the file is ~14 lines shorter and reads top-to-bottom without the rhetorical seam where the abandoned approach used to sit. Let me know if you'd rather keep the structural separation.

### 3. `histogram_plot()` `show_counts` + `y_relative` — documented

Added a one-liner to the `@param show_counts` docstring noting that labels stay as raw `n` even when `y_relative = TRUE`, with the rationale that raw counts read more naturally above proportion-scaled bars than the proportion itself would. Caller surprise prevented without changing behaviour.

### Out-of-scope commits also on this branch

Two extra commits landed here that aren't part of the dark-mode work — flagging so you can pull them out if you'd rather:

- `4b8112c` — `_quarto.yml` resource exclusions (`!.git/**`, `!node_modules/**`, `!renv/**`). Defensive against future resource-glob expansion accidentally walking those trees.
- `e26c781` — Parks a 246-line scratchpad prompt (`claude-code-translation-issues-prompt.md`) for an EN→FR translation pipeline. Pure WIP, not wired into the build.

Both were uncommitted in the working tree when the user asked to commit-and-push everything; I committed them separately so either is independently revertable (`git revert e26c781 4b8112c`) if you'd rather they live on `main` or a dedicated branch.

lddurbin:
All three follow-up items validated and addressed in `a39d28e`.

### Item 4 — `main-functions.R` line 9 "inline SVG" → "external SVG files"

Fixed. You're right that this matters more in the conventions block than it did in `setup.R` — the conventions block is the spec for future chart helpers, and getting the term wrong here would propagate. Took the opportunity to also spell out that Quarto wraps the file in `` so the next contributor knows what shape the output takes, not just what we call it.

### Item 5 — `_quarto.yml` no-op negation patterns

Removed. Verified your analysis: Quarto's `project.resources` is an additive list and the `!` prefix isn't honoured there. Even if it were, the only positive glob in this block (`assets/fonts/**`) couldn't have captured `.git/`, `node_modules/`, or `renv/`, so the negations gated nothing. They were defensive code masquerading as protection — exactly the false-confidence trap you flagged. Reverted to the original single-entry list.

If a future broader resource pattern (`**/*.png` style) becomes load-bearing, the right tool is Quarto's `project.exclude` key or a `.quartoignore` — not a `!`-prefixed entry in `resources`.

### Item 6 — `claude-code-translation-issues-prompt.md` at repo root

Moved to `docs/scratchpads/claude-code-translation-issues-prompt.md`.

I did first verify the leak risk before acting:

```
$ find _book -name "*.md"
(no output)
```

`CLAUDE.md` and `README.md` already live at the repo root and don't surface in `_book/`, so Quarto with the current config isn't auto-copying root `.md` files. Your concern is a hypothetical-but-fair one (someone later adds a wider glob and forgets that `.md` files exist at the root), so moving the scratchpad under `docs/scratchpads/` is cheap insurance even though no real leak exists today.

`docs/` already hosts ops-style documentation (`deviations-from-source.md`, `glossary-audit.md`, `pa11y-ignores.md`), so a `scratchpads/` subdirectory fits the existing conventions without minting a new top-level location.

---

That should close out the review thread. CI re-running now.

lddurbin:
Item 7 fixed in `bcec875` — replaced the `[[feedback_image_alt_lightbox]]` pointer with a self-contained one-liner noting the Quarto quirk and the figure-div workaround. You're right that the cross-reference was a leaky tooling-internal abstraction: it only resolves in a Claude Code session with this project's memory loaded, and would silently rot if the memory file is renamed.

Item 8 (filter selector scope is broader than ggplot2 outputs) — noted, not actioning this PR since nothing in the current diff triggers it, but agree the conventions block could use a sentence on this when the next non-ggplot SVG-emitting chunk lands. Tracking it mentally for the Day 3 figure work (#311–#316).

Will merge once CI clears.

PR #341, opened 2026-05-25
R figures: replace 7 Day 3 / Day 7 chart PNGs with R-generated SVGs (#311)

lddurbin:
Closes #311. Second card in the [#309](https://github.com/lddurbin/twelve_days_to_deming/issues/309) epic, picking up where [#310](https://github.com/lddurbin/twelve_days_to_deming/issues/310) (dark-mode-safe foundation) left off.

## Summary

- Replaces seven cropped-from-PDF chart PNGs with inline R/ggplot2 chunks across Day 3 (`run-chart-first-process`, `sales-control-chart`, `technical-aid-8`, favourite-example run + control) and Day 7 (`specification-limits`, `taguchi-loss-function` split into abstract parabola + Neave's personal graph).
- Extends `run_chart_plot()` with `central_line`, `y_label_fn`, `hline_label_side` (`"right"`/`"left"`), per-line `hline_label_offset`, `gridlines` (`"full"`/`"none"`) and `show_x_labels` so one helper covers both Neave's heavy-grid Technical Aid charts and his clean axis-only "favourite example" charts.
- Adds three new helpers — `taguchi_loss_abstract()`, `taguchi_loss_personal()`, `specification_limits_diagram()`.
- Introduces `R/data/` with the verbatim first-process 24-value sequence and the digitised favourite-example proportions; both decisions are recorded in [docs/deviations-from-source.md](docs/deviations-from-source.md).

## Two non-trivial data decisions (both logged in the deviations file)

- **Favourite-example proportions.** Neave does not print the underlying daily proportions — he himself redrew the run chart from Deming's *Out of the Crisis* account. The 31 values in [R/data/day-03-favourite-example.csv](R/data/day-03-favourite-example.csv) are a visual digitisation of Neave's printed chart on Day 3 page 21. The control limits and Central Line (LCL=3.1%, UCL=14.4%, CL=8.75%) are carried through verbatim.
- **Taguchi abstract parabola.** Neave's top panel has no scales — it's a *shape*, not a scaled function. The R version uses y = x² over [-1, 1] with the axis-cross and "nominal" label placed by eye to match the visual proportions of *DemDim* Figure 34. The bottom panel (Neave's personal graph) is data-faithful: the seven (temp, loss) readings printed on Day 7 page 21 are passed verbatim to `taguchi_loss_personal()`.

## Test plan

- [x] `quarto render` — full book, no chunk errors
- [x] All eight rendered SVGs visually compared against Neave's source crops
[...]

Issue #312, opened 2026-05-13
Day 3 histograms: Ford (x2) and two-histograms-24-values


lddurbin:
Tracking: #309

Depends on the foundation issue (specifically the `histogram_plot()` helper).

## Scope

| File | qmd |
|---|---|
| `assets/images/day-03/ford-histogram-with-compensation.png` | [content/days/day-03/02-at-the-ford-motor-company.qmd](content/days/day-03/02-at-the-ford-motor-company.qmd) |
| `assets/images/day-03/ford-histogram-without-compensation.png` | [content/days/day-03/02-at-the-ford-motor-company.qmd](content/days/day-03/02-at-the-ford-motor-company.qmd) |
| `assets/images/day-03/two-histograms-24-values.png` | [content/days/day-03/03-the-importance-of-time.qmd](content/days/day-03/03-the-importance-of-time.qmd) — note: this is the side-by-side "stacked boxes" vs conventional histogram comparison; needs a panelled layout |

## Out of scope

- `deming-histogram-caliper.png` — Deming's own figure from *Out of the Crisis*, kept as a quotation.

## Acceptance criteria

- [ ] Uses the `histogram_plot()` helper from foundation issue.
- [ ] Two-histograms figure uses `patchwork::wrap_plots()` or equivalent for the side-by-side layout.
- [ ] PNGs deleted.
- [ ] Deviations log updated.

PR #342, opened 2026-05-25
R figures: Day 3 histograms (Ford x2 + two-histograms-24-values) (#312)

lddurbin:
## Summary

Second child of epic #309 (after #311 landed in #341). Replaces three cropped-PNG histograms on Day 3 with R-generated SVGs that adapt to dark mode via #310's CSS-filter mechanism.

- Adds `stacked_boxes_plot()` (Day 3 page 7 left panel — stacked unit squares using `geom_point(shape = 22)` so the boxes stay square regardless of panel aspect)
- Adds `ford_histogram_plot()` (Day 3 page 5 — clean bars inside a shared LSL/USL knife-edge frame, no numeric axes — matches Neave's qualitative shape-illustration treatment)
- Composes the two-histograms-24-values panel via `patchwork::wrap_plots()` so both panels share one figure caption; reuses the 24-value first-process CSV already transcribed in #311

Data decisions and one alt-text correction (Neave's verbatim sequence has **five** 13s, not four as the source PNG caption claimed) are recorded in `docs/deviations-from-source.md`.

## Acceptance criteria (from #312)

- [x] Uses the `histogram_plot()` helper from #310 (right panel of two-histograms; Ford pair uses the sibling `ford_histogram_plot()` because the axes-less frame is qualitatively different).
- [x] Two-histograms figure uses `patchwork::wrap_plots()` for side-by-side layout.
- [x] PNGs deleted.
- [x] Deviations log updated.

## Test plan

- [x] `quarto render content/days/day-03/02-at-the-ford-motor-company.qmd` — clean
- [x] `quarto render content/days/day-03/03-the-importance-of-time.qmd` — clean
- [x] Visual check of all three rendered SVGs against Neave's source PDF — shape match
- [x] `./scripts/check-structure.sh 3` — 76/76 pass
- [x] `Rscript scripts/build-interday-audit.R` — no diff (no line numbers shifted that affect cross-references)
- [ ] CI Build/Structure/Audit/Accessibility green
- [ ] Production page renders cleanly in light + dark mode


lddurbin:
Thanks for the review — addressed both minor issues in `b39d44a`. Notes on the rest below.

**`library(patchwork)` placement** — fixed by moving the `library()` call into `R/functions/main-functions.R` alongside ggplot2/dplyr/gt/scales. It turned out to be more than a stylistic point: pa11y CI was failing on this PR with `there is no package called 'patchwork'`. Root cause was that `renv::snapshot()` ran *before* any source file used the package, so renv's source-scan silently omitted it from the lockfile — `renv::restore()` in CI then came up clean against an out-of-sync lockfile, and the qmd's inline `library(patchwork)` blew up at render time. Surfacing the import in main-functions.R both fixes the immediate failure and prevents the class of bug from recurring. Re-snapshotted; patchwork 1.3.2 is now recorded.

**Redundant `first_process` re-read** — fixed. The variable from the new `two-histograms-24-values` chunk is now relied on by the later `run-chart-first-process` chunk.

**`ford_histogram_plot()` baseline visibility** — confirmed correct as-is; the segment is intentionally drawn behind the bars so it's only visible in gaps and at the LSL/USL ends, matching Neave's printed appearance. No change.

**`box_size = 14` vs default 12** — fair point on the horizontal-crowding tradeoff; for the current `fig.width = 10` two-panel layout the boxes sit cleanly with visible gaps, and the function's docstring already covers the vertical-crowding tradeoff. Leaving as-is for now; if a future mobile-narrow rendering regresses we can either drop to the default 12 or expose a responsive override.

**Annotation text-size consistency between `stacked_boxes_plot()` and `ford_histogram_plot()`** — agreed it's a latent risk if they were ever composed side-by-side, but they aren't here. Logging it mentally for the next time a chart helper grows; if a future issue ends up composing both, that's the natural moment to factor the text-size constant out.

Issue #352, opened 2026-05-25
Optional Extras: rule-1-long / rule-1-short funnel illustrations


lddurbin:
Tracking: #314 (parent), #309 (epic)
Depends on #314a (`funnel_track_plot()` helper).

## Scope

Replace funnel rule illustrations in `assets/images/optional-extras/`:

- `rule-1-long.png`
- `rule-1-short.png`

## Implementation notes

- Direct reuse of `funnel_track_plot()` with different marker sequences/extents.
- Update the relevant Optional Extras qmd; identify file path in implementation.

## Acceptance criteria

- [ ] Both PNGs replaced with R chunks.
- [ ] Manifest (if applicable) + deviations log updated.

lddurbin:
Investigation update from automated implementation attempt (2026-05-27):

The agent that picked this up discovered that `rule-1-long.png` and `rule-1-short.png` are **not track diagrams** — they're control charts of Monte Carlo simulations under Rule 1 (rule-1-short: 15 simulated drops; rule-1-long: 50 drops, with UCL/CL/LCL drawn).

That means the original scope ("direct reuse of `funnel_track_plot()`") was based on a wrong assumption. The helper that's actually needed is either:

1. A funnel-rule *simulator* (Rule 1 dynamics → vector of resting positions), feeding the existing `run_chart_plot()` / control-chart machinery, or
2. A new `funnel_simulation_chart_plot()` helper that bundles simulation + chart in one call.

Either way this is meaningfully bigger than originally scoped and should not be folded into a same-size PR with #350/#351. Suggesting we:

- Keep this issue open as a placeholder
- File a new `#314x` dependency: *Add funnel-rule simulation helpers (Rules 1-4)*, blocking this issue
- Once that lands, this becomes a thin replacement-PR

No PR was created from the failed attempt — worktree was cleaned up automatically since no changes were committed.

Issue #348, opened 2026-05-25
Day 3 Funnel: dice-sequences visualisation (3.10)


lddurbin:
Tracking: #314 (parent), #309 (epic)

## Scope

Replace `assets/images/day-03/funnel-dice-sequences.png` on [content/days/day-03/10-introduction-to-the-funnel-experiment.qmd](content/days/day-03/10-introduction-to-the-funnel-experiment.qmd).

## Implementation notes

- Custom shapes needed: try `geom_text` with unicode dice glyphs (⚀⚁⚂⚃⚄⚅) first. Fall back to `ggimage` only if rendering proves unreliable across browsers / dark-mode invert.
- Single file change, single chart.

## Acceptance criteria

- [ ] PNG replaced with R chunk in Day 3.10.
- [ ] Manifest + deviations log updated.

PR #372, opened 2026-05-26
refactor: per-entry deviations log (one file per deviation)

lddurbin:
## Summary

- Move every deviations-log entry to its own file under `docs/deviations/` (22 files for the current log) plus a shared `_preamble.md`.
- Add `scripts/build-deviations-log.R` that assembles the stitched view from the directory.
- Keep `docs/deviations-from-source.md` in git as a build artifact (regenerated by the script) so external links — including the one in [`changes-from-source.qmd:37`](https://github.com/lddurbin/twelve_days_to_deming/blob/main/changes-from-source.qmd#L37) — keep working.

## Why

The old layout was one 71 KB Markdown monolith where every new entry was a top-of-file insertion. Two PRs landing in the same week reliably collided on the shared first line of that file — a *positional* conflict that had to be hand-resolved on every rebase (this happened four times during the #348–#352 agent-team run earlier today).

Splitting one entry per file turns the operation from a positional edit into an *additive* one. Concurrent PRs now touch different files and merge cleanly. The mechanical conflict that does remain is on the regenerated `docs/deviations-from-source.md`, but that's resolved by re-running the build script — much smaller than the positional-conflict tax it replaces.

This is the pattern that tools like `changesets`, `towncrier`, and `news-fragments` use for exactly the same reason in other projects.

## What's in this PR

- `docs/deviations/_preamble.md` — the intro / "Why this log exists" / "What counts" / "Entry format" sections.
- `docs/deviations/YYYY-MM-DD-slug.md` — 22 per-entry files, slugs derived from the original headings (capped at 60 chars).
- `docs/deviations/README.md` — contributor doc: how to add an entry, ordering rules.
- `scripts/build-deviations-log.R` — sorts entry files by filename descending, prepends the preamble, writes the stitched output with a "do not edit directly" banner.
- `docs/deviations-from-source.md` — regenerated. The diff against `main` is exactly the banner (+6 lines) plus same-date entries now strictly alphabetical-by-slug (which incidentally corrects the existing 2026-04-25 / 2026-04-26 ordering quirk between lines 213 and 222 of the old file).

## What's *not* in this PR

- No CI freshness check — if the user wants belt-and-braces, that can be a follow-up. The cost there is a small `.github/workflows/` addition that fails if `Rscript scripts/build-deviations-log.R` produces a non-empty `git diff`.
- No update to `changes-from-source.qmd:37` — the link target (`docs/deviations-from-source.md`) still exists.
- No memory / `feedback_deviations_log` update — that's user-level memory and will be updated once this lands so the convention shift is recorded.

## How to add a new entry after this lands

1. Create `docs/deviations/YYYY-MM-DD-short-slug.md` with the usual `## YYYY-MM-DD — title` heading + What/Where/Source/Why/Decided in/Landed in fields.
2. Run `Rscript scripts/build-deviations-log.R`.
3. Commit both the new entry file and the regenerated `docs/deviations-from-source.md`.

## Test plan

- [ ] CI passes (structure-check, inter-day audit, pa11y) — none of these workflows reference the deviations log, but verifying nothing else broke.
- [ ] Manual: clone this branch, run `Rscript scripts/build-deviations-log.R`, confirm `git diff docs/deviations-from-source.md` is empty (round-trip stable).
- [ ] Manual: confirm all 22 entry headings present in the regenerated file (already verified locally: `diff <(grep '^## 20' old) <(grep '^## 20' new)` is empty under sort).


lddurbin:
Thanks for the review. Both items addressed in 7378272:

1. **UTF-8 encoding** — `readLines` now passes `encoding = "UTF-8"` for both the preamble and entry files; output uses a file connection opened with `encoding = "UTF-8"`. Verified locally that the regenerated `docs/deviations-from-source.md` is byte-identical on macOS, so the change is a no-op on UTF-8 locales and a correctness fix on Windows.

2. **Opaque memory reference in README** — Rewrote the "Use today's date in NZ-local time" line to explain the convention inline (NZST/NZDT split, maintainer's location as the rationale) rather than delegating to `feedback_deviations_log`. A human contributor can now act on the instruction without external context.

Skipping the defensive entry-validation note since you flagged it as acceptable for an internal tool — agreed that adding heading-shape assertions would be gold-plating for a single-repo script that fails fast on a malformed entry anyway.

PR #377, opened 2026-05-27
feat: Optional Extras Part E PDFs + x-bar histograms (#358)

lddurbin:
## Summary

- Replaces the two cropped-PNG illustrations on Optional Extras printed pages 66–67 (Part E §3) with R/ggplot2 chunks. Page 66 (the four σ = 1 parent distributions used in Neave's first simulation: normal / uniform / triangular-unsymmetric / exponential) is composed as a 2×2 grid of small panels with corner labels, each auto-scaled to its own peak so the *shape* diversity reads directly. Page 67 (four histograms of the estimator *MR̄* ÷ 1.128 from simulated baselines of length 12) is the four-panel companion under a shared 0.0–2.5 x axis so the *spread similarity* claim of Neave's prose reads directly.
- Both chunks compose with `patchwork` inline rather than via the existing `pdf_family_plot()` helper. The helper is built around panels with shared scales (`overlay` or `stack` with fixed x/y), but Neave's page-66 figure *un-scales* per panel deliberately so the bell, the rectangle, the right-triangle and the spike all read at the same visual prominence. Widening the helper for a one-off bespoke layout would have added a wide surface area for a single caller — judgement call captured in the deviations log.
- Distribution parametrisations (all σ = 1): normal `N(0, 1)`; uniform on `[-√3, √3]` (variance `(2√3)²/12 = 1`); triangular right-triangle PDF on `[0, 3√2]` with mode at `0` (variance `(b-a)²/18 = 1`); exponential `rate = 1`. Neave does not give numeric `(min, mode, max)` for the triangular distribution anywhere in the text — only a shape illustration on Part D page 52 — so the shape is the load-bearing source-fidelity constraint and the endpoints follow from σ = 1.
- Simulation: N = 200,000 baselines of length 12 per distribution with `set.seed(358)`. Neave used 10,000,000; the "strikingly similar" visual claim reads obviously at 200K (peaks near 1.0 for the first three families, slightly wider with a small right shift for the exponential) and avoids slow per-render Quarto times.
- Both chunks carry `fig-cap` (verbatim from the original captions), `fig-alt` adapted for the new SVGs, `fig-width: 8` / `fig-height: 6`, and `#| lightbox: true`.
- Orphan PNGs `assets/images/optional-extras/part-e-p66-pdfs.png` and `…/part-e-p67-xbar-histograms.png` have been deleted (the conversion was straightforward to verify locally; no in-place keep-PNG-just-in-case pattern needed here).
[...]

Issue #354, opened 2026-05-25
Optional Extras (Part D): normal-family + area-percentages + pdf_family_plot() helper


lddurbin:
Tracking: #315 (parent), #309 (epic)

## Scope

Replace in `assets/images/optional-extras/`:

- `part-d-p47-normal-family.png` (family of normal curves at different σ)
- `part-d-p49-*.png` (area-percentages on a normal curve with shaded regions)

## Implementation notes

- Adds `pdf_family_plot()` helper in `R/functions/main-functions.R` — overlays multiple PDF curves (reused by #315f).
- Area-percentages charts: use `stat_function()` for the curve + `geom_ribbon()` for shaded tail regions.

## Acceptance criteria

- [ ] All listed PNGs replaced with R chunks.
- [ ] `pdf_family_plot()` helper available.
- [ ] Manifest (if applicable) + deviations log updated.

Issue #359, opened 2026-05-25
Optional Extras (Part E): subgroup illustrations (12 & 40)


lddurbin:
Tracking: #315 (parent), #309 (epic)

## Scope

Replace in `assets/images/optional-extras/`:

- `part-e-p69-12-subgroups.png` (12-subgroup illustration)
- `part-e-p70-40-subgroups.png` (40-subgroup illustration)

## Implementation notes

- Likely needs a `subgroups_strip_plot()` helper — small-multiple strip chart of subgroup means with subgroup boundaries marked. Confirm pattern against source before designing.
- Seeds documented in deviations log.

## Acceptance criteria

- [ ] Both PNGs replaced with R chunks.
- [ ] Helper added if newly designed.
- [ ] Manifest + deviations log updated.

PR #378, opened 2026-05-27
feat: Optional Extras Part D confidence intervals from R (#355)

### LINKED ISSUE #355: Optional Extras (Part D): confidence-interval illustrations
Tracking: #315 (parent), #309 (epic)

## Scope

Replace `part-d-p50-conf-*.png` in `assets/images/optional-extras/` — confidence interval illustrations.

## Implementation notes

- Adds `conf_interval_plot()` helper in `R/functions/main-functions.R` — normal curve + shaded central CI band + vertical line markers.
- Likely shares structure with #315b's area-percentages helper; consider whether to consolidate or keep distinct.

## Acceptance criteria

- [ ] All listed PNGs replaced with R chunks.
- [ ] `conf_interval_plot()` helper available.
- [ ] Manifest + deviations log updated.

lddurbin:
## Summary

- Replaces the two cropped-PNG illustrations on Optional Extras printed page 50 (the 95% and 99% confidence-region "yellow centre / red tails" bell-curve diagrams) with R/ggplot2 chunks. Both figures are rendered by a new reusable `conf_interval_plot(level, z, …)` helper added to `R/functions/main-functions.R`: it draws a standard-normal pdf, shades the central `level`-fraction yellow and the two tails red, drops vertical line markers at the boundaries, and labels the boundaries below the baseline as `μ±kσ` (matching Neave's printed labels).
- The helper is parameterised on `level` so a single call-site change is enough to produce any future centre-shaded confidence-region figure (e.g. the 99.8% / "3.09σ" remark Neave makes immediately after the page-50 figures). Boundary `z` values are passed explicitly as `1.96` and `2.58` in the chunks so the tick labels read verbatim as Neave's prose does; the helper's default (`z = qnorm((1 + level) / 2)`) is used only when a caller doesn't need to align with Neave's printed rounding.
- Standalone helper rather than extending the page-49 area-percentages helper from #354: the page-49 figure carries four band-percentages (34.1 / 13.6 / 2.15 / 0.135) annotated on each curve and expects three stacked panels; the page-50 figure is a single panel with a single yellow central region, red tails, and `μ±kσ` boundary labels. Extending the existing helper would have produced a wide-surface-area conditional function with mutually exclusive parameter branches; the two helpers stay separate and each reads clearly at its call-sites. See deviations log entry for the full rationale.
- Colour palette intentionally uses Neave's printed yellow (`#fff176`) for the central region and the project's existing red (`#ed0000`) for the tails. Neave's prose on page 50 introduces these specific colours by name ("a random variable has a 95% probability (coloured yellow) of lying within … 2.5% probability (left-hand tail coloured red)"), so the chromatic identity is pedagogically load-bearing.
- Both chunks are fully deterministic (closed-form `dnorm` + `seq()`), so no `set.seed()` is used — consistent with the PR #377 reviewer feedback that no-op seed calls should not appear in deterministic chunks. The new helper does not add `stats::` prefixes either, matching the project's convention for sibling helpers like `pdf_family_plot()`.
- Orphan PNGs `assets/images/optional-extras/part-d-p50-conf-95.png` and `part-d-p50-conf-99.png` are intentionally left in place per the epic #309 policy (delete only after the conversion has been verified in production — a follow-up PR removes them in batch).
- Deviations log entry added at `docs/deviations/2026-05-27-optional-extras-part-d-conf-intervals-from-r.md`; `docs/deviations-from-source.md` regenerated via `Rscript scripts/build-deviations-log.R`. Inter-day audit regenerated; no CSV changes.

Closes #355.

## Test plan

- [x] `Rscript -e 'source("R/setup.R")'` succeeds with the new helper present; `conf_interval_plot(0.95, z = 1.96)` and `conf_interval_plot(0.99, z = 2.58)` both return `ggplot` objects.
- [x] `Rscript -e 'testthat::test_dir("tests/testthat")'` passes 40/40 (5 new `conf_interval_plot` tests added on top of the existing 35).
- [x] `quarto render content/appendix/optional-extras/04-part-d-crash-course.qmd` succeeds; rendered HTML contains `` and `` wrapped in `<a class="lightbox">` elements with the full `fig-alt` text on the `` tags.
- [x] Visual spot-check confirms shading regions, boundary labels (`μ−1.96σ` / `μ+1.96σ` and `μ−2.58σ` / `μ+2.58σ`), and centred percentage labels ("95%", "99%") all match Neave's printed page 50.
- [ ] CI build of the full site succeeds.
- [ ] Reviewer confirms the 95% and 99% figures read as Neave's prose intends.


lddurbin:
Good catch on the scope creep. Confirmed and fixed in 718faf6 — the previous commit on this branch unintentionally bundled `xbar_false_signal_probs()` and `xbar_false_signal_panel()` into `R/functions/main-functions.R`. Those helpers belong to #382 (issue #359, Part E subgroups), and keeping them here would have caused a merge conflict with that branch.

Removed the 202-line Part E section cleanly. `conf_interval_plot()` and its 5 tests are unaffected — no other code in this PR's diff referenced the removed helpers.

On the other observations:

- **`yend = -Inf` and dead `if (n >= 2)` guard** — both lived in the now-removed Part E block. Will be picked up in #382 review (the bot has already flagged them there).
- **Grid precision for omitted `z`** — fair point. Both production call sites pass `z` explicitly, but I'll add a sentence to the `@param z` doc on the next touch so future callers know to pass an explicit `z` if they need pixel-precise boundaries.
- **Accessibility (yellow/red distinction)** — already documented as a known tradeoff in the deviations log; the `fig-alt` text carries the load for screen readers. Holding the current colour scheme since Neave names them explicitly in the surrounding prose.

Issue #396, opened 2026-05-28
Wire FR render into CI deploy workflow and extend smoke test


lddurbin:
**Context:** #320 commits us to keeping the existing single `rsync -avz --delete _book/` deploy. To honour that, the CI job must render both languages into a merged `_book/` before the rsync, and the smoke test must verify the FR side is healthy. This is the only sub-issue that touches CI and is the largest single piece of #321.

**Scope:**
- `.github/workflows/deploy.yml`: add a second render step (`quarto render --profile fr`) immediately after the existing `quarto render`, before the smoke test and rsync.
- Smoke test: assert `_book/fr/index.html` (or whatever the FR entry page is per the naming decision) exists. Keep the existing EN assertions intact. Day-by-day FR assertions stay degenerate at this stage (FR has only the stub) but the assertion **shape** should generalize so adding FR days later is a one-line change rather than a workflow refactor.
- `_freeze/` cache: confirm the existing GHA cache key (`renv.lock` + SHA, with restore-key fallbacks) captures FR-side `_freeze/` entries on save; no key change needed, but document the verification in the PR description.
- The merge_group triggers added in #385 should fire on this workflow too (they already do; just verify after edits).

**Acceptance Criteria:**
- [ ] `deploy.yml` runs both `quarto render` invocations sequentially before rsync.
- [ ] Smoke test fails if either `_book/index.html` or the FR entry page is missing.
- [ ] A green CI run on the PR shows EN and FR HTML files both produced and both shipped (verify via the workflow log).
- [ ] `_freeze/` cache hit rate for EN is unchanged across the before/after PR runs; FR contributes its own freeze entries on first run, captured by the same cache step.
- [ ] No change to the rsync line itself.

**Size:** S — one workflow file, parallel smoke assertions, plus a CI shakedown.

**Depends on:** the stub-page sub-issue.

Issue #394, opened 2026-05-28
Create FR root directory and stub French chapter


lddurbin:
**Context:** Validates the FR scaffold end-to-end by introducing one French page that renders through the FR profile and lands at the right URL. This is the first content commit on the FR side and the canonical example future FR chapters mirror.

**Scope:**
- Create the FR root directory (per the naming decision sub-issue).
- Add one minimal stub `.qmd` — `welcome.qmd` is the natural candidate. Body can be one paragraph (e.g., `Bienvenue — l'édition française est en cours de préparation.`).
- Point `_quarto-fr.yml`'s chapter list at the stub.
- The stub's Quarto section ID must match its EN sibling's ID so cross-references (#320 constraints section, "Cross-references") stay stable when translation lands.

**Acceptance Criteria:**
- [ ] The FR root directory exists with exactly one stub `.qmd`.
- [ ] `quarto render --profile fr` produces `_book/fr/<stub>.html` (and `_book/fr/index.html` if the stub is the entry page) with `<html lang="fr">`.
- [ ] The stub's `@sec-…` ID matches the EN sibling's `@sec-…` ID.
- [ ] EN build still produces only `_book/<en-files>` — no new files appear outside `_book/fr/`.

**Size:** XS — one new content file, one config edit.

**Depends on:** the profile file sub-issue.

Issue #320, opened 2026-05-23
Multilingual architecture decision (spike)


lddurbin:
**Context:** Quarto has no first-class multi-language book mode, so EN/FR coexistence is an architectural choice that constrains everything downstream.

**Scope:** Evaluate Quarto profiles vs. a parallel `fr/` directory vs. a separate branch; decide site structure (e.g. `/fr/`) and switcher behaviour; document the trade-offs and the chosen approach. Decision and documentation only — no production scaffolding.

**Acceptance Criteria:**
- [ ] The options are documented with trade-offs (in `workflow/`).
- [ ] One approach is chosen and recorded with rationale.
- [ ] The decision specifies where French sources live and how the build targets them.

**Size:** S — config/documentation, no production code; roughly half a day to one day.

**Depends on:** none

PR #399, opened 2026-05-28
feat(quarto): add _quarto-fr.yml profile for French build

### LINKED ISSUE #393: Add _quarto-fr.yml profile
**Context:** #320 specifies that the French build is driven by a Quarto profile activated via `quarto render --profile fr`, with `lang: fr` and `output-dir: _book/fr`. This ticket creates the profile file. The stub French page that gets rendered is added separately (next sub-issue).

**Scope:**
- Add `_quarto-fr.yml` adjacent to `_quarto.yml`.
- Override only what differs for FR: `lang: fr`, `output-dir: _book/fr`, and a `chapters:` list pointing at the FR stub chapter (added next).
- Inherit theme, CSP `header-includes`, page footer, filters, and the reading-prefs script from `_quarto.yml`.
- Do **not** introduce FR-specific assets, scripts, or page-footer copy in this PR — those are post-scaffold work.

**Acceptance Criteria:**
- [ ] `_quarto-fr.yml` exists with `lang: fr`, `output-dir: _book/fr`, and a chapter list with at least one entry.
- [ ] `quarto render` (no `--profile`) still produces the EN book under `_book/` with no regressions in sidebar, page footer, or smoke-test output count.
- [ ] `quarto render --profile fr` runs to completion locally (one stub page is fine).
- [ ] `_book/fr/` does not appear in EN builds.

**Size:** XS — one new YAML file inheriting most settings.

**Depends on:** #320, plus the naming decision sub-issue.

lddurbin:
## Summary

Adds the Quarto profile file activated by `quarto render --profile fr`, per the architectural decision in #320 (Option A). The profile is intentionally minimal — overrides only `lang`, `project.output-dir`, `book.chapters`, and `book.appendices`; everything else (theme, CSP header-includes, page-footer, filters, format settings) inherits from `_quarto.yml`.

The chapter list references `content-fr/welcome.qmd`, which is added by the parallel PR for #394. Until that lands, `quarto render --profile fr` will error on the missing source — expected at this stage of the scaffold. The CI deploy workflow is not modified here; FR rendering is wired into CI by the later #396 PR.

## Test plan

- [x] YAML parses (validated via `ruby -ryaml`).
- [x] `_quarto.yml` untouched (no EN regression possible from this PR alone).
- [x] Profile overrides only what differs; no theme/CSP/filter redeclaration.
- [ ] Once #394 merges: `quarto render --profile fr` produces `_book/fr/welcome.html`.

Closes #393

### DIFF
--- a/_quarto-fr.yml
+++ b/_quarto-fr.yml
@@ -1,0 +1,9 @@
+lang: fr
+
+project:
+  output-dir: _book/fr
+
+book:
+  chapters:
+    - content-fr/welcome.qmd
+  appendices: []


lddurbin:
Thanks — all three notes are non-actionable in this PR, but one is worth locking in:

**`site-url` inheritance is intentional and permanent.** The architecture decision in #320 explicitly commits the project to subpath-only deployment — EN at `/`, FR at `/fr/`, both rendered into one `_book/` and shipped by the existing single rsync to `deming.leedurbin.co.nz`. There is no planned future state in which FR moves to a subdomain, so `site-url` will not need overriding in any subsequent i18n PR. (If that changes, the rollback would be a `_quarto-fr.yml` line plus a deploy-target change in `.github/workflows/deploy.yml`, both trivial — but it's not on the roadmap.)

The other two points (render-time verification of `appendices: []` and `book.title` localisation) are correctly identified as later concerns: `appendices: []` gets its first real exercise once #394 + #396 land, and `book.title` localisation belongs with the actual FR-content work rather than the scaffold.

PR #406, opened 2026-05-28
Language switcher: link EN <-> FR editions on every page (#322)

### LINKED ISSUE #322: Language switcher
**Context:** Readers need to move between the two editions.

**Scope:** Add a navbar (or equivalent) switcher linking corresponding EN/FR pages; handle the case where a page exists in one language only.

**Acceptance Criteria:**
- [ ] A working switcher links the two editions from every page.
- [ ] Missing-translation cases degrade gracefully (no broken links).

**Size:** S — one focused UI area, limited surface; roughly half a day to one day.

**Depends on:** #321

lddurbin:
## Summary

Adds a language switcher present on **every page of both editions** that links the reader to the corresponding page in the other language (EN ↔ FR). Closes #322.

## Approach

- **Where it lives:** a small script, `assets/scripts/language-switcher.js`, referenced from `header-includes` in the base `_quarto.yml`. Because both the EN (default) and FR profiles inherit that shared base, the script runs on every page of both editions. The book project type has no per-page navbar slot, so `header-includes` is the one hook guaranteed everywhere — and it's exactly how the existing reading-prefs toggle is wired.
- **The UI:** a fixed `<a>` in the bottom-**left** corner (the reading-prefs toggle owns the bottom-right, so they never overlap). It mirrors that toggle's visual language: same border/shadow, dark-mode palette, high-contrast and reduced-motion handling, 44px min hit target, icon-only collapse on narrow viewports, and `is-yielding` fade when the page-navigation footer scrolls into view.
- **The link:** target resolved client-side from the current path. Visible label is the destination-language autonym (`Français` / `English`) carrying `lang`/`hreflang`; `aria-label` is written in the *current* page's language (`Read this page in French` / `Lire cette page en anglais`).

## Graceful-degradation rule (no broken links)

EN and FR pages are **not** a clean path-prefix mirror — EN content lives under `/content/...` with home at `/index.html`, FR content under `/fr/content-fr/...` with home at `/fr/index.fr.html`. So the EN↔FR correspondence is an explicit `PAGE_MAP` in the script. Resolution:

1. If the current page has a known counterpart in `PAGE_MAP` → link to it.
2. Otherwise (the common case today, FR being a stub) → link to the **other edition's home page**.

Either way the target is a path the build is known to emit, so the switcher **cannot 404**. As FR pages are translated, adding their EN↔FR pair to `PAGE_MAP` upgrades the link from home-fallback to same-page automatically — no other change needed.

## Acceptance criteria + evidence

- [x] **A working switcher links the two editions from every page.**
  - Script referenced on **all 121** EN HTML pages and both real FR content pages (`index.fr.html`, `content-fr/welcome.html`). The 3rd FR file is the redirect stub `fr/index.html`, which needs no script.
  - `language-switcher.js` is copied into both `_book/assets/scripts/` and `_book/fr/assets/scripts/`; Quarto relativizes the `src` per page depth (e.g. `../../../assets/...` on deep EN pages, `../assets/...` on the FR welcome page).
- [x] **Missing-translation cases degrade gracefully (no broken links).**
  - The switcher can only ever emit 4 targets: `/index.html`, `/welcome.html`, `/fr/index.fr.html`, `/fr/content-fr/welcome.html` — all verified to be real files in the built `_book/` tree.
  - Simulated resolution across representative pages: EN deep pages with no FR counterpart (e.g. `day-03/01-...`, `accessibility`) correctly fall back to `/fr/index.fr.html`; EN home/welcome map to their exact FR counterparts; FR home/welcome map back to their exact EN counterparts; trailing-slash `/fr/` and `/` normalise to the right home pages.

## Verification

- `quarto render` (EN, 121 pages) then `quarto render --profile fr` (FR) both succeed — EN first so its output-dir clean doesn't wipe nested `_book/fr`.
- `node --check` on the script and YAML validation on both profiles pass.
- Accessibility: rendered as a real link with `href`, `aria-label`, `lang`/`hreflang`; keyboard-operable by default; honours dark mode, `prefers-contrast: high`, `prefers-reduced-motion`, and print (hidden).

### DIFF
--- a/_quarto.yml
+++ b/_quarto.yml
@@ -47,6 +47,7 @@ format:
         })();
       </script>
       <script src="/assets/scripts/reading-prefs.js"></script>
+      <script src="/assets/scripts/language-switcher.js"></script>
       <script>
         document.addEventListener("DOMContentLoaded", function () {
           var a = document.createElement("a");
--- a/assets/scripts/language-switcher.js
+++ b/assets/scripts/language-switcher.js
@@ -88,7 +88,7 @@
     link.setAttribute("hreflang", destLang);
     link.setAttribute("aria-label", ariaLabel);
     link.innerHTML =
-      '<span class="lang-switcher-icon" aria-hidden="true">\u{1F310}</span>' +
+      '<span class="lang-switcher-icon" aria-hidden="true">🌐</span>' +
       '<span class="lang-switcher-label">' + destLabelText + "</span>";
 
     document.body.appendChild(link);
--- a/assets/scripts/language-switcher.js
+++ b/assets/scripts/language-switcher.js
@@ -1,0 +1,116 @@
+/* Language switcher (issue #322).
+ *
+ * Renders a fixed, keyboard-operable control on every page of both editions
+ * that links the reader to the corresponding page in the other language.
+ *
+ * It is a sibling affordance to the reading-preferences toggle
+ * (assets/scripts/reading-prefs.js): same fixed-corner placement language,
+ * same focus-ring and dark-mode conventions, same "inject once at
+ * DOMContentLoaded" lifecycle. It is rendered by JS rather than baked into the
+ * Quarto sidebar because the book project type has no per-page navbar slot,
+ * and header-includes (shared by both profiles in _quarto.yml) is the one
+ * place guaranteed to run on every page of both editions.
+ *
+ * GRACEFUL DEGRADATION RULE
+ * -------------------------
+ * EN and FR pages are NOT a clean path-prefix mirror: EN content lives under
+ * /content/... with the home page at /index.html, while FR content lives under
+ * /fr/content-fr/... with the home page at /fr/index.fr.html. So the EN<->FR
+ * correspondence is expressed explicitly in PAGE_MAP below (EN path -> FR
+ * path). The link target is resolved as:
+ *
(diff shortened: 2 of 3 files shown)


lddurbin:
Both addressed in 6b0b77c — `npm test` (55) green, `node --check` clean.

**1. Keyboard focus on the yielding switcher — fixed.** Added `visibility: hidden` to `.lang-switcher.is-yielding`, which drops the off-screen invisible link from the tab order (WCAG 2.4.7 / 2.4.11). Confirmed it doesn't interfere with the existing `opacity`/`transform` transition (`visibility` isn't in that transition list).

I also took your note on the pre-existing `.reading-prefs-toggle.is-yielding` and applied the same one-line fix there — it's the identical defect one block away and cheap to close, so it seemed worth keeping the two sibling controls consistent rather than leaving a known (if milder) gap. The `.reading-prefs-panel.is-yielding` needed nothing: it already uses `display: none`, which is out of the tab order.

**2. ES6 brace escape — fixed.** Replaced `\u{1F310}` with the literal `🌐` glyph. I went with the literal rather than the `🌐` surrogate pair to match the established style in `reading-prefs.js`, which places its icon glyph (`Aa`) directly in the `innerHTML` string; the source is UTF-8 so the glyph round-trips fine. Either way the ES2015 syntax is gone.

Thanks — both were good catches.

Issue #331, opened 2026-05-23
Segment alignment and divergence classification


lddurbin:
**Context:** Align candidates per segment and classify disagreement by type so mechanical issues auto-fix and only genuine prose calls reach the reviewer.

**Scope:** Align candidates from #328/#329 per segment using the back-translation signal from #330; classify agreement (full/partial/split); tag each divergence `glossary`, `register`, or `prose` with a short reason; emit a structured per-segment record for synthesis.

**Acceptance Criteria:**
- [ ] Every segment has an agreement classification.
- [ ] Each divergence is typed (glossary/register/prose).
- [ ] Glossary/register divergences are separated from prose divergences.

**Size:** M — alignment logic plus classification heuristics applied across the full corpus; the dividing line between glossary/register/prose needs care; roughly one to three days.

**Depends on:** #328, #329, #330

lddurbin:
Parking this pending a French-speaking collaborator.

The FR edition is blocked on a human reviewer, not on tooling. Every step from here — MT/LLM candidates, divergence classification, reinjection, parity checks — produces output that nobody on the project can currently adjudicate, so running the pipeline would just accumulate unverifiable French text against a course whose whole value is fidelity to Neave's source.

**What's already built and stays built:** segment extraction (#323, #324, #325) and the glossary + validation lock (#326, #327) are complete and merged. Reviving the edition starts from working extraction and an approved termbase, not from zero.

**What revives this:** a French speaker with enough quality-management background to adjudicate Deming terminology. At that point reopen this milestone as a group — the dependency chain between these nine is intact and still correct.

Closing as not planned rather than done: none of this work happened, and none of it was rejected on merit.

PR #407, opened 2026-05-28
feat(i18n): prose extractor + reinjector with identity round-trip (#323)

### LINKED ISSUE #323: Prose extraction, reinjection, and identity round-trip
**Context:** The technical crux. A non-destructive parser that pulls translatable Markdown prose from `.qmd` and writes it back into the exact structural position, preserving everything else.

**Scope:** Extract prose segments with a stable ID and structural address; **exclude and preserve** code chunk bodies, inline code, cross-references (`@sec-`/`@fig-`/`@tbl-`), citation keys (`[@…]`), shortcodes, fenced-div attributes, maths, and YAML keys (only `title`/`subtitle`-type values are translatable); build the inverse reinjection step; provide an identity round-trip test (extract, reinject the *original* strings, diff against source).

**Note on this codebase:** the parser must preserve the `#sec-pageN` anchors and relative `../day-NN/*.qmd#sec-pageN` link forms documented in `workflow/PATTERNS.md`, plus the `<dfn id="…">` glossary anchors at first defining instance, plus the existing `<span lang="ja|de|fr|la|es">` non-English wrappers. None of these are translatable; all must survive byte-identical.

**Acceptance Criteria:**
- [ ] Prose segments carry a stable ID and structural address.
- [ ] Code, refs, citations, shortcodes, maths, and YAML keys are never emitted as translatable.
- [ ] **Identity round-trip produces byte-identical files for every `.qmd`** (must pass before any translation runs).

**Size:** M — self-contained tooling component with clear deliverable and a hard correctness gate (byte-identical round-trip); roughly one to three days.

**Depends on:** none (foundational; after reading house conventions)

lddurbin:
## What & why

Foundational component of the EN→FR translation pipeline (per `workflow/multilingual-architecture.md`, Option A). A **non-destructive** parser that pulls translatable Markdown prose out of `.qmd` files and writes translated text back into the exact structural position, leaving everything else byte-for-byte untouched.

## Design

A `.qmd` is treated as an ordered sequence of **lines**. Each line is classified as **structural** (preserved verbatim) or part of a **translatable prose block**; contiguous prose lines are grouped into *segments*.

Each segment carries:
- a **stable id** — `path::kind#ordinal::xxhash64(text)`, content-derived (not coupled to absolute line numbers of unrelated edits);
- a **structural address** — `{file, kind, start_line, end_line}`;
- the raw `text` (exact source bytes);
- a `masked` form + `placeholders` map, in which every inline construct that must not reach a translator is replaced by an opaque PUA-delimited sentinel and restored losslessly.

**Byte-faithfulness is structural, not best-effort:** reinjection splices replacement lines into each segment's line range and reproduces every structural region from the original file. Reinjecting a segment's own text therefore reproduces the source exactly — the identity round-trip is correct by construction (including files with no trailing newline; the corpus is LF-only and 10 files lack a final newline — all handled).

### Excluded / preserved (never emitted as translatable)
- YAML front matter — only the **values** of `title` / `subtitle` / `description` are extracted; keys and every other field are structural. The masked surface for a title is the bare value (the `title: "` prefix and closing quote are placeholders).
- Fenced **code chunks** (```` ``` ````/```` ```{...} ````/`~~~`), bodies **and** fences.
- Block-level **HTML** structural tags on their own line (`<div>`, `</div>`, `<details>`, `<figure>`, `<table>`, …).
- Fenced-div fences (`:::`/`::::`), bare anchors (`[]{#sec-pageN}`), blanks, HTML comments.
- Inline (masked within prose): inline code, `@sec-`/`@fig-`/`@tbl-` refs, `[@citations]`, shortcodes `{{< … >}}`, inline & display maths.

### Repo-specific preservation (per `workflow/PATTERNS.md`)
- `#sec-pageN` anchors and inline `[]{#sec-pageN}` attachments.
- Relative `../day-NN/*.qmd#sec-pageN` **and** `../days/day-NN/*.qmd#sec-pageN` link forms.
- `<dfn id="…">` glossary anchors (whole element masked).
- `<span lang="…">` non-English wrappers (whole element masked; `<span class=…>` deliberately **not** treated as a lang wrapper).

## Acceptance criteria

- [x] **Prose segments carry a stable ID and structural address** — id = `path::kind#ord::hash`; address = `{file,kind,start_line,end_line}`. Verified stable across repeated extraction (test) and addresses well-formed.
- [x] **Code, refs, citations, shortcodes, maths, and YAML keys are never emitted as translatable** — dedicated masking with a placeholder map; unit tests assert each construct is absent from the masked surface and present in `placeholders`; code-chunk bodies excluded entirely.
- [x] **Identity round-trip is byte-identical for every `.qmd`** — **122 / 122 files byte-identical**, 4778 prose segments extracted. Gate: `Rscript R/translation/roundtrip.R` → `PASS: all files byte-identical.`

### Evidence
```
$ Rscript R/translation/roundtrip.R
Identity round-trip: 122/122 files byte-identical (4778 prose segments extracted).
PASS: all files byte-identical.

$ Rscript -e 'testthat::test_dir("tests/testthat")'
[ FAIL 0 | WARN 0 | SKIP 0 | PASS 328 ]   # 235 new assertions in test-prose-extract.R
```

## Files
- `R/translation/prose-extract.R` — extractor, reinjector, mask/unmask, corpus enumerator.
- `R/translation/roundtrip.R` — CLI identity gate (optional `--json <path>` writes the full extraction sidecar).
- `tests/testthat/test-prose-extract.R` — full-corpus round-trip + per-construct masking + mask/unmask losslessness + reinjection-isolation.

No new dependencies (`yaml`, `jsonlite`, `digest`, `testthat`, `here` already in `renv.lock`); no `renv::snapshot()` needed.

## Notes for the reviewer / limitations
- **Segment granularity is block-level prose**, not sentence-level — appropriate for the foundational stage; sentence alignment is #331's job.
- **Inter-day link text is conservatively preserved** inside the masked link token (the whole `[text](../day-NN/…#sec-pageN)` is masked). This honours #323's "link forms must survive byte-identical" literally; if downstream wants the link *label* translatable, that's a deliberate follow-up, not a regression.
- The line-classifier assumes LF endings and simple one-line YAML scalars — both verified true for the entire current corpus (no CRLF, no block scalars). A block-scalar `title: |` would currently be treated as structural rather than extracted; none exist today.
- Pure tooling addition, no source-content departure, so no `docs/deviations-from-source.md` entry.

Closes #323

### DIFF
--- a/R/translation/prose-extract.R
+++ b/R/translation/prose-extract.R
@@ -1,0 +1,474 @@
+# prose-extract.R
+#
+# Foundational tooling for the French-translation pipeline (issue #323).
+#
+# A NON-DESTRUCTIVE prose extractor + reinjector for Quarto `.qmd` files.
+#
+# Design contract
+# ---------------
+# A `.qmd` file is treated as an ordered sequence of LINES. Each line is
+# classified as either STRUCTURAL (preserved verbatim, never translated) or
+# part of a TRANSLATABLE PROSE block. Contiguous translatable lines are grouped
+# into prose *segments*, each carrying:
+#   * a stable `id`            (content-derived hash, order-tagged for stability)
+#   * a structural `address`   (region kind + start/end line, byte offsets)
+#   * the raw `text`           (exact source bytes of the segment)
+#   * a `masked` form + `placeholders` map, in which every inline construct that
+#     must NOT be exposed to a translator (inline code, cross-refs, citations,
+#     shortcodes, maths, <span lang>, <dfn id>, #sec-pageN anchors, relative
+#     ../day-NN/*.qmd#sec-pageN links) is replaced by an opaque token. The mask
+#     is what a translator/MT engine consumes; the raw text is what the identity
+#     round-trip reinjects.
+#
+# What is STRUCTURAL (never emitted as translatable prose):
+#   * YAML front matter, except the VALUES of translatable keys
+#     (title / subtitle / description) which become their own prose segments.
+#   * Fenced code blocks (``` / ```{...} / ~~~), bodies AND fences.
+#   * Lines that are purely an HTML structural tag (<div ...>, </div>,
+#     <details>, <summary> alone, <button ...>, </button>, <figure>, <table> …).
+#   * Fenced-div fences (::: / :::: with attributes).
+#   * Bare span/anchor-only lines such as `[]{#sec-page7}`.
+#   * Blank lines, separators, HTML comments.
+#
+# Reinjection is the exact inverse: a translated value is provided per segment id;
+# every structural region is reproduced byte-for-byte from the original. The
+# IDENTITY round-trip (reinject each segment's own raw text) therefore yields a
+# byte-identical file by construction. That is the hard correctness gate.
+#
+# Dependencies: base R + yaml + jsonlite + digest. (stringr/purrr deliberately
+# avoided so the tool runs under a minimal renv subset.)
+
+suppressPackageStartupMessages({
+  library(yaml)
+  library(jsonlite)
+  library(digest)
+})
+
+# ---------------------------------------------------------------------------
+# Low-level IO: read a file as its exact byte string, and split into lines
+# while remembering whether the final line had a trailing newline.
+# ---------------------------------------------------------------------------
+
+#' Read a file's full content as a single UTF-8 string (exact bytes).
+.read_raw_text <- function(path) {
+  sz <- file.info(path)$size
+  if (is.na(sz)) stop("cannot stat file: ", path)
+  raw <- readBin(path, "raw", n = sz)
+  txt <- rawToChar(raw)
+  Encoding(txt) <- "UTF-8"
+  txt
+}
+
+#' Split content into physical lines, recording the trailing-newline flag.
+#' Returns list(lines = character vector WITHOUT newline chars,
+#'              trailing_newline = logical).
+#' Note: the corpus is LF-only (verified); we assume "\n" line endings.
+.split_lines <- function(content) {
+  if (identical(content, "")) {
+    return(list(lines = character(0), trailing_newline = FALSE))
+  }
+  trailing <- grepl("\n$", content)
+  # strsplit drops a trailing empty field, so a trailing "\n" yields the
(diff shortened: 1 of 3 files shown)


lddurbin:
Thanks for the thorough review. Addressed in c470cb8 — all 339 testthat assertions pass and the identity round-trip stays 122/122 byte-identical.

**1. Token collision — respectfully, a false positive (but I improved the code anyway).** The inline tokens were *already* PUA-delimited: the source was `sprintf("%d", counter)`, not `sprintf("%d", counter)`. The U+E000/U+E001 sentinels are raw bytes embedded in the string literal, so a diff viewer renders them invisibly and the line reads as a bare integer — which is exactly what it looked like in review. Your own example round-trips correctly on the committed code:

- `"Score 1 point. Use \`foo\` to check 2 things."` → unmask → identical ✅

I've still acted on the underlying readability problem: inline tokens now use the shared `.ph_token()` helper (built from the named `` constants), so the line is clean ASCII, self-documenting, and unified with the YAML token scheme — no more invisible bytes to mislead the next reader. I also added your suggested digit-collision test as a permanent regression guard (it passes, confirming no collision).

**2. YAML descending range — fixed.** Switched to `seq_len(yaml_end - 2L) + 1L` so empty front matter yields an empty range instead of `c(2, 1)`.

**3. `.is_comment_line` — dead code removed; `-->$` kept deliberately.** Dropped the `&& FALSE` clause (agreed, dead). I did **not** tighten `-->$` to `^<!--`, though: that clause is load-bearing. It catches the *closing* line of a multi-line comment block (opened by `<!--` on an earlier line) so both ends break the surrounding prose block at the call site (line ~385). Requiring `^<!--` would reclassify every comment-closing line as translatable prose. Added a comment explaining the intent.

**4. `<summary>`/`<button>` raw tags — fixed properly.** Rather than just commenting, added an `html_label_tag` pattern that masks the wrapper tags (open/close) while leaving the inner label on the translator surface, so an MT/LLM step never sees the raw HTML. New test covers it.

**5. Test gap — added** (the collision guard above, plus the wrapper-tag test).

Issue #325, opened 2026-05-23
Extract UI and ARIA strings from ObservableJS / HTML / JS


lddurbin:
**Context:** The interactive layer in this site — the Funnel Experiment (`assets/scripts/funnel-experiment.js`, with SVG `aria-labelledby`/`aria-describedby`, live regions, button text), the cooperation tables (`assets/scripts/cooperation-tables.js`), the in-page text inputs (`Inputs.textarea`, `Inputs.text` placeholders in `.qmd` OJS chunks), the "Click this after writing your comments" disclosure buttons, the chapter download buttons, and clock indicators — contains button text, input labels, live-region messages, and ARIA labels that must be translated for accessibility parity.

**Scope:** Extend the extractor to capture `ui-string` and `aria-label` segments across the interactive elements; carry interface context so they can be translated as UI text; reuse the #323 reinjection path; extend the round-trip test.

**Acceptance Criteria:**
- [ ] UI and ARIA strings are extracted as distinct types with interface context.
- [ ] JS/HTML logic and element structure are preserved exactly.
- [ ] Identity round-trip remains byte-identical with UI/ARIA strings included.

**Size:** M — multiple file types (OJS in `.qmd`, raw HTML divs with `role`/`aria-label`, plain JS), each with different parsing concerns; roughly one to three days.

**Depends on:** #323

Issue #324, opened 2026-05-23
Extract user-facing strings from R chunks


lddurbin:
**Context:** R chunks (via the helpers in `R/functions/main-functions.R` — `run_chart_plot`, `red_beads_control_chart`, `render_redbeads_table`, `create_clock`) generate user-facing plot titles, axis/legend labels, table headers and captions that need translating; the surrounding R code does not.

**Scope:** Extend the extractor from #323 to identify translatable string literals in R chunks, tagged `r-string`; reuse the #323 reinjection path; extend the round-trip test to cover them.

**Acceptance Criteria:**
- [ ] User-facing R strings are extracted as `r-string`, distinct from prose.
- [ ] R code logic is never altered; only string literals are touched.
- [ ] Identity round-trip remains byte-identical with R strings included.

**Size:** S — extension of an existing extractor with a narrow surface (helpers in `R/functions/`); roughly half a day to one day.

**Depends on:** #323

lddurbin:
Implemented and merged in PR #410 (squash `0c4f016`). User-facing R strings now extract as `r-string` segments from both `.qmd` R chunks and the `R/functions/` helper defaults, via a tight label-position whitelist; identity round-trip stays byte-identical (125/125 at merge, 129/129 after #325 added the JS corpus). `render_redbeads_table` excluded by design (its strings are CSS/data keys, not labels) — see the deviation log.

PR #410, opened 2026-05-29
feat(translation): extract user-facing R-chunk/helper strings as r-string segments (#324)

lddurbin:
## Summary

Implements issue #324. Extends the #323 prose extractor so **user-facing string literals in R code** — chart titles, axis/legend labels, on-canvas annotations, and the chart-helper *default* labels — are extracted as a distinct segment kind **`r-string`**, separate from `prose`.

The three hard gates hold:
1. User-facing R strings are extracted as a distinct `r-string` kind.
2. R code logic is **never** altered — only the bytes *inside* a whitelisted label literal become translatable.
3. The identity round-trip stays **byte-identical** with R strings included.

## Approach

- **New reusable module `R/translation/code-string-extract.R`** (base-R only), sourced by `prose-extract.R`. It provides language-agnostic primitives so the sibling issue #325 can plug OJS/JS in via the same seams:
  - `.tokenize_string_literals()` — a small hand-rolled quote/escape scanner that returns each literal's character span.
  - `.scan_code_line_literals(line, in_func_params)` — returns only the literals at whitelisted, label-bearing positions as sub-line span records.
  - `.update_func_param_context()` — a cross-line paren tracker so a helper default-arg label is recognised even though its `function(` opener is several lines above (the signatures are multi-line).
  - `.reinject_subline_segments()` — **generalized intra-line reinjection** that splices a replacement into a literal's `start_col..end_col` span, leaving every other byte of the line byte-identical. Works for any sub-line kind, not just `r-string`.
  - `.mask_code_string()` — masks R escapes (`\n`, `\t`, …), `sprintf`/`printf` format specifiers (`%s`, `%d`, …) and `glue` `{...}` interpolation off the translator surface, reusing #323's PUA-sentinel placeholder scheme.
- **`prose-extract.R`** now: scans `` ```{r} `` chunk bodies for whitelisted literals; adds `extract_r_file()` + `r_corpus()` (scoped to `R/functions/`) so hardcoded helper-default labels are captured; gives `r-string` segments `start_col`/`end_col` addressing; and makes `reinject_qmd()` apply sub-line edits (intra-line) before whole-line edits.
- **`roundtrip.R`** gate now covers the `.R` helper corpus in addition to the `.qmd` corpus.

## Intra-line reinjection design

`r-string` segments are **sub-line**: a literal living inside one line (e.g. the `Shaft diameter` inside `labs(x = "Shaft diameter")`). Their `address` carries 1-based character offsets (`start_col`/`end_col`, the inner span between the quotes) with `start_line == end_line`. Reinjection applies sub-line splices **first**, on the original line numbering (descending column order within a line so earlier offsets stay valid), then the existing whole-line prose/yaml splicing. Sub-line segments only ever sit inside code chunks while whole-line segments sit outside, so they never contend for the same line.

## Whitelist — exactly which R string positions are extracted, and why

A deliberately **tight** whitelist (when in doubt, EXCLUDE — a false positive that rewrote a data key/colour/path would corrupt logic):

| Position | Why translatable |
|---|---|
| `ggtitle(<first positional>)` | the plot title |
| `labs(title= / subtitle= / caption= / x= / y= / fill= / colour= / color= )` | axis/legend/title labels |
| `annotate("text", label = ...)` | on-canvas text labels — **gated on the `"text"` geom**; `annotate("segment"/"point", …)` is excluded |
| helper default args `lsl_label`, `usl_label`, `x_title`, `y_label`, `x_label`, `legend_title`, `percent_label`, `caption`, `subtitle`, `title` inside a `function(...)` parameter list | the hardcoded default labels in `R/functions/main-functions.R` |

**Explicitly NOT extracted** (verified by tests): file paths (`here::here("R/data/...")`), hex colours / aesthetic values (`fill = "magenta"`, `"#ff0000"`), `gt(rowname_col = "Name")` and other data keys, `factor(levels = …)`, `tab_options`/`theme` styling tokens (`"white"`, `"bottom"`), and any literal in a non-R chunk (`ojs`, `python`).

Variable-valued defaults (`fill_colour = CHART_LINE_COLOUR`, `labs(y = y_label)`) are naturally skipped because they are not string literals.

## Deviation / architectural note

`render_redbeads_table` was **excluded**: inspection shows it contains no user-facing label literals — every string is a CSS/styling value (`"white"`, `"solid"`, `"bottom"`) or a gt data/row key (`"Name"`, `"Totals"`, `"Daily Totals"` as a `rows=` selector). Extracting any of these would corrupt styling or row selection. Logged in `docs/deviations/2026-05-29-extract-user-facing-r-chunk-strings-as-r-string-segments.md`.

## Results

- **Round-trip:** `Rscript R/translation/roundtrip.R` → `125/125 files byte-identical (4802 segments extracted). PASS`.
- **Tests:** `testthat::test_dir("tests/testthat")` → all green, including the existing `prose-extract` suite (untouched) and the new `r-string-extract` suite covering: whitelisted literal extracted; non-user-facing literals excluded; intra-line reinjection rewrites only the literal; multi-literal line splicing; escape masking; `.R` + `.qmd` identity round-trips.


lddurbin:
Thanks for the careful review. Addressed in the latest push (round-trip still 125/125 byte-identical; r-string suite still green):

**1. Redundant `code_is_r` branch — fixed.** Confirmed: `}` is already in the `[ ,}]` class, so branch two was unreachable. Collapsed to a single `grepl` and folded in your tab suggestion (`[ \t,}]`) so tab-separated chunk options match too.

**2. R 4.0+ raw strings — documented, but no runtime guard (intentional).** I traced the actual exposure and it's already safe on two independent counts, so a guard would be gold-plating:
- The whitelist matcher only fires when a `(` or `name =` sits *immediately* before the opening quote. A raw string puts `r`/`R` there instead (e.g. `ggtitle(r"(...)")` → `pre` ends in `r`, not `(`), so the positional/named matchers never accept it — a raw-string literal is never emitted as a segment.
- The identity round-trip reinjects original bytes, so byte-identity holds even where the tokenizer mis-spans a raw string.

I added a comment in `.tokenize_string_literals()` recording the limitation and pointing the next maintainer at where to add an `r"`/`R"` prefix guard *if* a future label position ever needs raw strings.

**3. `reinject_qmd` handling both `.qmd` and `.R` — documented.** Added a docstring note that it's the single reinjection path for both file kinds (whole-line + sub-line for `.qmd`, sub-line only for `.R`).

**4. Hardcoded line indices in tests — fixed.** Both assertions now select the code line by a stable token (`grep("ggtitle"/"labs(")`) instead of `[[1]][6]`, so they survive any front-matter line-count change in the fixture.

PR #411, opened 2026-05-29
feat(translation): extract UI/ARIA strings from OJS, JS and raw-HTML (#325)

### LINKED ISSUE #325: Extract UI and ARIA strings from ObservableJS / HTML / JS

lddurbin:
Closes #325. Extends the #323 prose / #324 r-string translation pipeline to extract user-facing INTERFACE strings off three new surfaces, as two distinct segment kinds that carry interface context.

## Approach

Builds entirely on #324's reusable sub-line machinery:
- `.tokenize_string_literals()` now recognises the JS **backtick** delimiter (its `\`-escape scan already closes backtick literals correctly).
- `.mask_code_string()` gained a `js_interp` pattern (`\$\{[^}]*\}`) ordered **before** the existing glue `{…}` pattern, so a `${…}` interpolation is masked as one opaque PUA-sentinel token — never leaving a stray `$` or exposing live code to a translator. The raw segment `text` keeps the exact interpolation bytes for byte-identical reinjection.
- `.scan_code_line_literals()` is the generic seam; a sibling `.scan_ojs_js_line_literals()` applies the OJS/JS whitelist and returns a per-literal **kind**.
- New whitelist tables drive `.ojs_js_literal_kind()`; a raw-HTML attribute scanner (`.scan_html_attr_values()`) runs before the structural-skip; a cross-line `.update_announce_call_context()` catches multi-line `announceFunnelStatus(...)` messages.
- `extract_js_file()` + `js_corpus()` added (mirroring #324's `extract_r_file`/`r_corpus`, using the per-kind `next_ord` convention per the #324 follow-up note). `reinject_qmd()` flows the new sub-line kinds through unchanged. The round-trip gate and a new test extend to the `.js` corpus.
- Each new segment carries an optional `context` hint (`ojs-input` / `html-attr` / `js`) inside its `address`, so the top-level segment schema (`id`/`address`/`text`/`masked`/`placeholders`) is identical to existing segments (AC #1 interface context).

## Whitelist positions (exact)

**OJS chunks** — `label:` / `placeholder:` values of `Inputs.textarea` / `Inputs.text` / `Inputs.button` → `ui-string`.

**JS files** (`assets/scripts/*.js`):
- RHS string literal of `X.textContent =` / `X.placeholder =` → `ui-string`
- VALUE arg of `setAttribute("aria-label"|"placeholder"|"title", <literal>)` → `aria-label`/`ui-string` (gated on the attribute-NAME arg)
- first string-literal arg of `announceFunnelStatus(...)` (incl. multi-line form) → `aria-label`

**Raw-HTML attributes in `.qmd`** — `aria-label` → `aria-label`; `title`/`alt`/`placeholder` → `ui-string`. Scanned on structural tag lines before they are skipped.

## Deliberately excluded

`role=`/`class=`/`id=`/`data-*=`/`aria-controls=`/`aria-live=`/`aria-atomic=`/`tabindex=`/`headers=`; ARIA role/live token values (`"status"`, `"polite"`, `"true"`, `"button"`); element-id templates (`` `track-title-${id}` ``); CSS class lists (`"fe-sr-live visually-hidden"`); selectors and event names; file paths/filenames (incl. `createDownloadButton(..., "file.txt")`); non-literal RHS (variable/call/concatenation — the matcher only fires when a quote/backtick is flush after the position); equality tests (`=== "empty"`, excluded by an assignment lookbehind); and EMPTY string literals (`cell.textContent = ""`). `<summary>`/`<button>` inner label text is already captured by #323's prose path and is **not** double-handled.

## SVG `<title>`/`<desc>` decision — SCOPED OUT (documented deviation)

`funnel-experiment.js` builds its SVG accessibility text by concatenation (`` svg += `<title id="${titleId}">Funnel Experiment Track</title>` ``). The translatable words sit inside a larger template literal mixing SVG markup + interpolated ids; there is no clean call/assignment position to whitelist, and extracting the whole literal would put `<title id="…">` markup on the translator surface and risk corrupting the SVG. The two strings remain English; a follow-up can hoist them into a `textContent`/`setAttribute` position the whitelist already covers. The ternary-branch consts (`"No stage active."`, `"Dark"`/`"Light"` toggle states) are excluded on the same rationale. Logged in `docs/deviations/2026-05-29-extract-ui-aria-strings-from-ojs-js-html.md`.

## Results

- **Round-trip**: `Rscript R/translation/roundtrip.R` → **129/129 files byte-identical (5458 segments)**, PASS.
- **Tests**: `Rscript -e 'testthat::test_dir("tests/testthat")'` → **FAIL 0 | WARN 0 | SKIP 0 | PASS 445** (existing prose + r-string suites stay green; new `test-ui-string-extract.R` covers AC (a)–(f)).
- Live extraction sample: OJS labels/placeholders, JS `Download Your Notes`/`Download Your Exercise`/`Area ${…}:` (interpolation masked), the multi-line funnel announce message (6 `${…}` masked), and raw-HTML `aria-label` values (`Commentary`, `Thought`, `Author's note`, …) all captured; no `role`/`class`/`id` false positives.


lddurbin:
Good catch on the real bug — fixed in the latest push (round-trip still 129/129 byte-identical; suites green).

**Bug: `\b` matched inside `data-aria-label` — fixed.** Confirmed your analysis empirically: `\baria-label` matches at the `-`→`a` boundary in `data-aria-label`, and the live extractor was pulling the `data-aria-label`/`data-title` values. Replaced the `\b` prefix with a negative lookbehind `(?<![A-Za-z0-9-])` (your second suggested form), which rejects any preceding word char *or* hyphen — so it excludes `data-aria-label`, `data-title`, `data-alt`, etc., while still matching the real `aria-label`/`title` on the same tag. Verified the regression case now extracts only `Good`, dropping `BUG`. Also corrected the misleading comment that claimed `\b` handled this.

**Test gap — closed.** Added a dedicated regression test (`a whitelisted attribute name as the TAIL of a data-* attribute is NOT extracted`): a tag with `data-aria-label`, `data-title`, plus a real `aria-label` and `title`, asserting the `data-*` values are excluded from both kinds while the real ones extract. The pre-existing `data-bs-target` case didn't catch this because it shares no substring with a whitelisted name, as you noted.

**Minor: compound-assignment operators — fixed.** Extended the assignment lookbehind to `(?<![=<>!+*/%&|^-])` so `+= -= *= /= %= &= |= ^=` are excluded alongside the comparison operators. An append like `x.textContent += "…"` is a fragment, not the full UI text, so it should not be extracted — even though it's absent from the current corpus and the round-trip would preserve bytes regardless.

lddurbin:
Thanks — both addressed in the latest push (round-trip 129/129; suites green).

**Compound-assignment test gap — closed.** Added `cell.textContent += "Append fragment";` to the (d) exclusion block, so the guard now has a regression test. As you noted, an append is a fragment rather than the full label, and if the lookbehind is ever narrowed this case will fail.

**`??=` — documented, deliberately not excluded.** I left the behaviour as-is and added a comment explaining why, because excluding it would likely be *wrong*: a nullish-coalescing default (`x.textContent ??= "Default fallback"`) is usually a legitimate user-facing fallback string we'd *want* to extract, not a false positive — and it's absent from all four `.js` files anyway. Worth noting the related logical-assignment forms `&&=`/`||=` are already excluded for free, since their last char before `=` is `&`/`|`, both in the exclusion class; only `??=` (with `?`) falls outside it. Flagged in the comment for the next contributor who touches the whitelist.

I'll treat the JS-operator edge cases as settled here — the corpus has none of them and the round-trip preserves bytes regardless, so further hardening would be speculative. Merging once CI is green.

PR #417, opened 2026-05-29
feat(translation): add glossary corpus loader (flattened segment frame)

### LINKED ISSUE #412: Add glossary corpus loader (flattened segment frame)
**Context:** Decomposed from #326. Three glossary-seeding tasks (matcher, candidate discovery, assembler) all need to read the three extracted corpora into one tidy place. Extracting that read path as its own module — built and merged first — is what lets the rest fan out safely (the #324/#325 shared-core lesson: sequence the shared core, then parallelise the disjoint work).

**Scope:** New `R/translation/glossary-corpus.R` exposing `glossary_segments(root = ".")`. It runs the existing `qmd_corpus()` / `r_corpus()` / `js_corpus()` extractors (see `R/translation/prose-extract.R` and `roundtrip.R`) and returns a single tidy data frame, one row per segment, with columns:

- `text` — the segment text
- `kind` — `prose` / `yaml` / `r-string` / `ui-string` / `aria-label`
- `occurrence_type` — `prose` / `r` / `ui`, derived from `kind` (+ `address.context`)
- `file`, `start_line`

Pure read. No term logic, no matching, no French — those land in the sibling issues.

**Acceptance Criteria:**
- [ ] `glossary_segments()` returns one row per extracted segment across all three corpora.
- [ ] `occurrence_type` mapping (kind/context → prose|r|ui) is covered by tests.
- [ ] Output is deterministic / idempotent on re-run.
- [ ] Dependencies stay base R + yaml + jsonlite (no new packages).

**Size:** XS — single new file reusing existing corpus extractors.

**Depends on:** #323, #324, #325 (all merged). **Shared core — merge before #326c/#326d/#326e.**

Part of #326.

lddurbin:
Closes #412

## Summary

Adds `R/translation/glossary-corpus.R` exposing a single function `glossary_segments(root = ".")` — a **pure-read** flattener that runs the existing extractor's three corpus discovery + extraction paths and produces ONE tidy base `data.frame`, one row per extracted segment, across all three corpora:

- `.qmd`: `qmd_corpus()` -> `extract_qmd()`
- `.R`:   `r_corpus()`   -> `extract_r_file()`
- `.js`:  `js_corpus()`  -> `extract_js_file()`

### Output columns (exactly)
`text`, `kind`, `occurrence_type`, `file`, `start_line`.

### Mapping
- The extractor emits `kind == "yaml_value"` for front-matter scalars; this is normalised to `"yaml"` in the `kind` column.
- `occurrence_type` is derived from `kind`: `prose|yaml -> "prose"`, `r-string -> "r"`, `ui-string|aria-label -> "ui"`. The mapping lives in small, directly-testable helpers (`.normalise_kind`, `.occurrence_type`). `address$context` is accepted by the helper but deliberately unused — `kind` alone partitions the space unambiguously (documented in a comment).

### Constraints honoured
- No term logic, no matching, no French (sibling issues).
- Deterministic / idempotent: corpora iterated in qmd, r, js order; files via the existing sorted `*_corpus()` helpers; segments in extraction order.
- Base R + `yaml` + `jsonlite` only (transitively `digest` via the extractor). No dplyr/tibble — returns a base `data.frame` with `stringsAsFactors = FALSE`.

## Tests
`tests/testthat/test-glossary-corpus.R` (style mirrors `test-prose-extract.R`):
- row count == sum of `extract_*` segment counts across all three corpora, all five columns present + correctly typed;
- direct unit tests of the kind/context -> occurrence_type mapping for **every** kind value (plus an unmapped-kind error guard);
- frame-level consistency (no raw `yaml_value` leaks; `occurrence_type` agrees with `kind`);
- idempotence (`identical()` on two calls).

## Verification
`Rscript -e 'testthat::test_file("tests/testthat/test-glossary-corpus.R")'` -> **PASS 30 / FAIL 0**.

Smoke run of `glossary_segments(".")`: **5458 rows**.
- occurrence_type: prose 4778, ui 656, r 24
- kind: prose 4656, yaml 122, aria-label 81, ui-string 575, r-string 24


lddurbin:
Thanks for the review. Actioned and pushed in cd859ed:

**✅ Dead `context` vector — fixed.** Removed the `character(n)` allocation and its loop population from `.segments_to_rows()`; the call site is now `.occurrence_type(raw_kind)`. Kept the `context = NULL` parameter on `.occurrence_type()` itself since the block comment there documents the deliberate design decision. 30/30 tests still pass.

**↩︎ Per-file error handling — declining for now.** `glossary_segments()` is a pure-read helper over a version-controlled corpus that the full test suite already exercises end-to-end (the identity round-trip gate in `R/translation/roundtrip.R` would catch a malformed source file long before this loader runs). Wrapping each `extract_*` in `tryCatch` to re-throw with a filename adds a layer of indirection for a failure mode the upstream gate already localises. Happy to revisit if the loader ever runs over untrusted/partial inputs.

**↩︎ Self-locating IIFE — no change.** Agreed it's non-trivial, but it's a deliberate mirror of the same block in `prose-extract.R` so the two stay consistent; diverging here would be worse than the shared complexity. Noted the `sys.frames()` fragility for if the sourcing mechanism ever changes.

PR #418, opened 2026-05-29
data(glossary): seed SoPK/experiment termbase CSV (#413)

### LINKED ISSUE #413: Curate SoPK/experiment termbase seed CSV
**Context:** Decomposed from #326. Deming terminology has established French renderings — *Système de Connaissance Approfondie*, *causes communes / causes spéciales*, PDSA, the red bead and funnel experiments. **Seed from established sources; do not invent.** This issue is data + sourcing only (no code), so it is disjoint from #326a and can run in parallel.

**Scope:** Hand-author `glossary/seed-terms.csv` covering, at minimum:
- the 7 `<dfn>` terms (common-cause, special-cause, in-statistical-control, operational-definition, pdsa-cycle, system-of-profound-knowledge, transformation — see `docs/glossary-audit.md`);
- named experiments / frameworks: Red Beads, Funnel, Six Processes, 14 Points, Seven Deadly Diseases, Joiner Triangle, and the four parts of the SoPK.

Columns: `term_en`, `fr_rendering`, `source` (citation to established French Deming literature), `aliases` (pipe-separated matching variants, e.g. `common cause|common-cause variation`), `decision_needed`.

**Acceptance Criteria:**
- [ ] Every `<dfn>` slug and every named experiment/framework above is present.
- [ ] Each row has a sourced `fr_rendering` **or** is flagged `decision_needed = TRUE`.
- [ ] No rendering is invented without a `source`; contested ones are flagged.

**Size:** XS — a single curated CSV, no tooling.

**Depends on:** — (none). Pairs with #326a in wave 1.

Part of #326.

lddurbin:
Closes #413

## Summary

Adds `glossary/seed-terms.csv` — a sourced seed termbase for the SoPK / named-experiment terminology, ahead of the FR translation consume work. Data and sourcing only; no code or tooling.

Columns (exact): `term_en, fr_rendering, source, aliases, decision_needed`.

### Coverage (17 rows)
- The 7 `<dfn>` glossary terms (slugs from `docs/glossary-audit.md`): common-cause, special-cause, in-statistical-control, operational-definition, pdsa-cycle, system-of-profound-knowledge, transformation.
- Named experiments/frameworks: Red Beads, Funnel, Six Processes, 14 Points, Seven Deadly Diseases, Joiner Triangle.
- The four parts of the System of Profound Knowledge as their own rows: Appreciation for a System, Theory of Variation, Theory of Knowledge, Psychology.

### Sourcing
French renderings are seeded only from established francophone Deming sources — primarily the Association Française Edwards Deming (fr-deming.org) and Jean-Marie Gogue's translations of *Out of the Crisis* (*Hors de la crise*) and the 1989 Versailles lecture *La connaissance approfondie*. Every row carries a concrete citation in `source`.

No rendering is invented without a source. Uncertain/contested renderings are flagged `decision_needed = TRUE` rather than guessed:
- **pdsa-cycle** — Deming disavowed PDCA/"roue de Deming"; no normalised FR translation of the PDSA acronym exists.
- **six-processes** — no established FR rendering found for the named framework.
- **joiner-triangle** — no established FR rendering found (term itself was a rejected `<dfn>` candidate).

### Verification
- `Rscript -e 'read.csv(...)'`: parses to 17 rows × 5 columns, correct headers.
- Python `csv.DictReader`: all `decision_needed` values are TRUE/FALSE; no empty `fr_rendering` lacks a flag; every row has a non-empty `source`; UTF-8 accents intact.
- decision_needed = TRUE: 3 rows (pdsa-cycle, six-processes, joiner-triangle); FALSE: 14.


lddurbin:
Thanks for the careful review. I validated each point against the file; this one is data-only so I'm being conservative about churning it before the consume-side matcher (#326c) defines its contract.

**↩︎ Trailing blank line (line 19) — not present.** The file is 18 lines (1 header + 17 data) and ends with a single terminating newline, no phantom record:
```
$ wc -l glossary/seed-terms.csv  →  18
$ python3 -c "import csv;rows=list(csv.reader(open('glossary/seed-terms.csv')));print(len(rows));print([i for i,r in enumerate(rows) if all(c=='' for c in r)])"
18
[]   # zero empty/phantom rows
```
`od -c` confirms the last bytes are `...,FALSE\n` — a standard POSIX line terminator, not a blank row. No change needed.

**↩︎ Redundant `Psychology|psychology` and `PDSA` aliases — keeping, intentionally.** Whether these are redundant depends entirely on the matcher's case-folding policy, which #326c hasn't defined yet. Under a **case-sensitive** matcher `Psychology` and `psychology` are distinct, useful keys; only a case-insensitive matcher makes them duplicates. Seeding both is a deliberate hedge so the seed data doesn't silently constrain the matcher design. Same logic for the `PDSA` alias vs `fr_rendering`. I'd rather dedup these *once the matcher fixes the contract* than guess now.

**↩︎ Alias/boolean convention undocumented — agree, deferring to the consume side.** A README/CSVW descriptor is the right call, but the formal contract (pipe separator, TRUE/FALSE, what `aliases` feeds) is owned by the matcher/assembler issues (#326c–#326e), not this seed-data PR. Capturing it here would risk documenting a contract that those issues then change. Tracking it for that wave.

**↩︎ `out of statistical control` antonym alias — intentional.** Per `docs/glossary-audit.md` the `in-statistical-control` entry deliberately covers both poles ("also called stable/predictable; the opposite, out of statistical control, means…"), so a lookup on either resolves to the same term entry. Good call that the intent isn't obvious from the data alone — an explicit `antonyms` column is worth considering when the matcher schema is designed.

Net: no changes to the CSV — point 1 doesn't reproduce, and 2–5 are matcher-contract decisions I'd rather not pre-empt from the seed layer. Happy to fold any of them in if you'd prefer them locked now.

PR #427, opened 2026-06-25
fix(prefs): hide language switcher until FR edition is ready

lddurbin:
## Summary

- Removes the Language row from the reading-preferences panel (`reading-prefs.js`) so users don't see a broken EN↔FR switcher before the French edition is live
- Cleans up the now-unused language infrastructure: `PAGE_MAP`, `EN_HOME`, `FR_HOME`, `FR_TO_EN`, `currentIsFr`, `normaliseLangPath`, `languageTarget`
- Updates `README.md` to accurately describe the preferences panel (it has both dark mode and dyslexia font, not just a font toggle), adds the missing high-contrast support entry to the accessibility section, and removes the stale mention of the language switcher

The language row will be re-introduced in the FR-launch PR once the French edition is ready to deploy.

## Test plan

- [ ] Open the live site (or `quarto preview`) and confirm the `Aa` preferences panel shows only **Theme** and **Dyslexia font** rows — no Language row
- [ ] Confirm Theme and Dyslexia font toggles still work correctly and persist across page navigation
- [ ] Confirm README accessibility section reads accurately for promotion purposes

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ Each day includes:
 - **Responsive Design**: Optimized for desktop, tablet, and mobile devices
 - **Search and Navigation**: Easy content discovery and cross-referencing
 - **Print-Friendly**: Clean, formatted output for printing or PDF generation
-- **Accessibility**: Skip-link, optional dyslexia-friendly font, ARIA-labelled interactive elements, and reduced-motion support — see the [Accessibility](#-accessibility) section below for the full list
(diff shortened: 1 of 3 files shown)


lddurbin:
Good catch — agreed. Removed all four locations: the base block (`.reading-prefs-lang-link` / `.reading-prefs-lang-arrow`), the dark-mode overrides, and the two compound selectors in the high-contrast and dark+high-contrast blocks. Pushed as a follow-up commit.

